PR #1, opened 2016-07-27
Add cluster interfaces

mmihic:
Intentionally modeled on consul APIs
(https://godoc.org/github.com/hashicorp/consul/api)

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,13 +20,17 @@
 
 package cluster
 
+import (
+	"github.com/golang/protobuf/proto"
+)
+
 // KVStore provides access to the configuration store
 type KVStore interface {
 	// Get retrieves the value for the given key, unmarshalling it into the provided value
-	Get(key string, v interface{}) error
+	Get(key string, v proto.Message) error
 
 	// Put stores the value for the given key
-	Put(key string, v interface{}) error
+	Put(key string, v proto.Message) error
 }
 
 // A ServiceInstance is a single instance of a service
--- a/client.go
+++ b/client.go
@@ -22,7 +22,7 @@ package cluster
 
 // KVStore provides access to the configuration store
 type KVStore interface {
-	// Get retrieves the value for the given key, marshalling it into the provided value
+	// Get retrieves the value for the given key, unmarshalling it into the provided value
 	Get(key string, v interface{}) error
 
 	// Put stores the value for the given key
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -0,0 +1,77 @@
...
+	Advertise(ad Advertisement) error
+
+	// Unadvertise indicates a given instance is no longer available
+	Unadvertise(service, id string) error
+
+	// QueryInstances returns the list of available instances for a given service
+	QueryInstances(service string, opts *QueryOptions) ([]*ServiceInstance, error)
+
+	// QueryEndpoints returns a list of available endpoints for a given service
+	QueryEndpoints(service string, opts *QueryOptions) ([]string, error)
```
mmihic: This method exists to provide a pool-friendly approach to retrieving all of the endpoint strings for the service.  May be redundant.

client.go
```diff
@@ -0,0 +1,77 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cluster
+
+// KVStore provides access to the configuration store
+type KVStore interface {
+	// Get retrieves the value for the given key, marshalling it into the provided value
```
xichen2020: `unmarshaling it into the provided value` perhaps, assuming we are just getting a blob from the KVStore?

client.go
```diff
@@ -0,0 +1,77 @@
...
+	Get(key string, v interface{}) error
+
+	// Put stores the value for the given key
+	Put(key string, v interface{}) error
+}
+
+// A ServiceInstance is a single instance of a service
+type ServiceInstance struct {
```
xichen2020: Do we need rack information here as well?

robskillington: Can we perhaps make all the structs simply `interface`s with accessors e.g. "Service() string" and where mutable with setters too "SetHealth(fn func() error)"?

Have found that when you depend on a third party vendor "A" and they include struct "X" while another third party "B" itself depends on "A" which includes struct "X" it tends to cause a lot of issues.  This is why we couldn't use Bark in M3DB because one of our other dependencies had a dependency on Bark too and said that the structs in Bark were being declared multiple times.

Also everything being interfaces does make test relatively less painful as it's very easy to generate a mock.

mmihic: both sgtm

client.go
mmihic: Yes, meant to be unmarshalling.  Unlike more general purpose config APIs, I want this layer to be prescriptive about how data objects are stored (specifically as protobufs) so they will explicitly handling marshalling / unamarshalling.

client.go
mmihic: That said, I'd dispute that everything being interfaces makes testing less painful - now you need to mock non-behavioral value objects, where as if they were just structs you could just use them.

FWIW our mocking approach is really unmaintainable.  We need a deterministic build step that doesn't require developers to know exactly what mocks exists and how we are mapping interfaces into mock files.

client.go
```diff
@@ -0,0 +1,77 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cluster
+
+// KVStore provides access to the configuration store
+type KVStore interface {
+	// Get retrieves the value for the given key, marshalling it into the provided value
+	Get(key string, v interface{}) error
```
robskillington: If we are ok with deciding now that we will go the protobuf route we can make these of type `proto.Message`: https://godoc.org/github.com/golang/protobuf/proto#Message

Or we could make a new interface:
```go
type KVStoreValue interface {
  Marshall() ([]byte, error)
  Unmarshall(data []byte) error
}
```

We would have to create some container for protobufs though as `proto.Message` does not implement this.  We could:

```
type ProtobufValue interface {
  KVStoreValue
  Get() proto.Message
}

type protobufValue struct {
  value proto.Message
}

func NewProtobufValue(v proto.Message) {
  return &protobufValue{v}
}

func (v *protobufValue) Marshall() ([]byte, error) {
  return proto.Marshall(v.value)
}

func (v *protobufValue) Unmarshall(data []byte) error {
  return proto.Unmarshall(data, v.value)
}
```

Then use as such:
```
foo := NewThingProtobufObject() // type *ThingProtobufObject implements proto.Message
foo.innerKey = "innerValue"

kv := NewKVStore(...)

kv.Put("foo", NewProtobufValue(myProtobufValue))

bar := NewThingProtobufObject() // type *ThingProtobufObject implements proto.Message
err := kv.Get("bar", NewProtobufValue(bar))
// if err == nil then bar is returned value
```

PR #2, opened 2016-08-03
placement algo and service

cw9:
@xichen2020 @robskillington @mmihic


placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
+package algo
```
mmihic: Needs license header (see the other files in this repo)

placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
...
+	return ph.generatePlacement(), nil
+}
+
+func (a placementAlgorithm) AddHost(ps placement.Snapshot, addingHost placement.Host) (placement.Snapshot, error) {
+	ph, addingHostShards := newAddHostPlacementHelper(ps, addingHost)
+	targetLoad := ph.hostHeap.getTargetLoadForHost(addingHostShards.hostAddress())
+	// try to steal shards from the most loaded hosts until the adding host reaches target load
+outer:
+	for len(addingHostShards.shardsSet) < targetLoad {
+		if ph.hostHeap.Len() > 0 {
```
mmihic: short circuit and eliminate the else

````go
if ph.hostLeap.Len() == 0 {
   return nil, errCouldNotReachTargetLoad
}

tryHost := heap.Pop(ph.postHeap).(*hostShards)
//... etc

````

File: placement/algo/algo.go [hunk 3]
```diff
@@ -0,0 +1,450 @@
...
+			for shard := range tryHost.shardsSet {
+				if ph.canAssignHost(shard, tryHost, addingHostShards) {
+					ph.assignShardToHost(shard, addingHostShards)
+					ph.removeShardFromHost(shard, tryHost)
+					heap.Push(ph.hostHeap, tryHost)
+					continue outer
+				}
+			}
+		} else {
+			return nil, errors.New("new host could not reach target load")
```
mmihic: nit: use global err var

placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
...
+			ph.removeShardFromHost(shard, leavingHostShards)
+
+			addingHostShards.addShard(shard)
+		} else {
+			shardsUnAssigned = append(shardsUnAssigned, shard)
+		}
+	}
+	// if there are shards that can not be moved to adding host
+	// distribute them to the cluster
+	err := ph.placeShards(shardsUnAssigned, leavingHostShards)
```
mmihic: nit:

``go
if err := ph.placeShards(shardsUnassigned, leavingHostShards); err != nil
``

mmihic: nit: ```shardsUnassigned``` (only capitalize at the start of a word, unassigned is its own word)

placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
...
+
+type placementAlgorithm struct {
+}
+
+func newM3DBPlacementAlgorithm() placement.Algorithm {
+	return placementAlgorithm{}
+}
+
+func newM3DBDeploymentPlanner() placement.DeploymentPlanner {
+	return placementAlgorithm{}
```
mmihic: Seems weird that these are the same (empty) type

placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
...
+
+var (
+	errNotEnoughRacks = errors.New("not enough host to take shards, please make sure RF <= number of racks")
+	errHostAbsent     = errors.New("could not remove or replace a host that does not exist")
+)
+
+type placementAlgorithm struct {
+}
+
+func newM3DBPlacementAlgorithm() placement.Algorithm {
```
mmihic: Would prefer not to use M3DB terminology here

placement/algo/algo.go
```diff
@@ -0,0 +1,450 @@
...
+	return placementAlgorithm{}
+}
+
+func newM3DBDeploymentPlanner() placement.DeploymentPlanner {
+	return placementAlgorithm{}
+}
+
+func (a placementAlgorithm) InitPlacement(hosts []placement.Host, ids []int) (placement.Snapshot, error) {
```
mmihic: Usually ``Init`` is used to initialize an object with state.  I think in this case it is building the initial placement? If so I'd call it ``BuildInitialPlacement`` or something similar.

placement/algo/algo.go
cw9: yeah will create an empty deploymentPlanner type

placement/algo/algo.go
cw9: done

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+import (
+	"container/heap"
+	"errors"
+	"math"
+
+	"github.com/m3db/m3cluster/placement"
+)
+
+var (
+	errNotEnoughRacks          = errors.New("not enough host to take shards, please make sure RF <= number of racks")
```
xichen2020: Perhaps this should be "not enough racks to take shards"?

placement/interface.go
```diff
@@ -0,0 +1,58 @@
...
+	// HostsLen returns the length of all hosts in the placement
+	HostsLen() int
+
+	// Replicas returns the replication factor in the placement
+	Replicas() int
+
+	// ShardsLen returns the number of shards in a replica
+	ShardsLen() int
+
+	// ShardIDs returns all the unique shard ids for a replica
```
xichen2020: `Shards returns all...`

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+)
+
+type rackAwarePlacementAlgorithm struct {
+}
+
+func newRackAwarePlacementAlgorithm() placement.Algorithm {
+	return rackAwarePlacementAlgorithm{}
+}
+
+func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Host, ids []int) (placement.Snapshot, error) {
```
xichen2020: maybe call it `shards` instead of `ids` to be consistent with the interface?

placement/interface.go
```diff
@@ -0,0 +1,58 @@
...
+	Shards() []int
+}
+
+// HostShards represents a host and its assigned shards
+type HostShards interface {
+	Host() Host
+	Shards() []int
```
robskillington: We have been using `uint32` everywhere for shards, perhaps we stick with this?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+func (hh *hostHeap) Push(h interface{}) {
+	host := h.(*hostShards)
+	hh.hosts = append(hh.hosts, host)
+}
+
+func (hh *hostHeap) Pop() interface{} {
+	n := len(hh.hosts)
+	host := hh.hosts[n-1]
+	hh.hosts = hh.hosts[0 : n-1]
+	return host
```
robskillington: Perhaps we should take into account the case where heap is empty and return `nil` in this case?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+
+func (ph placementHelper) generatePlacement() placementSnapshot {
+	return placementSnapshot{uniqueShards: ph.uniqueShards, rf: ph.rf, shardsLen: ph.getShardLen(), hostShards: ph.hostShards}
+}
+
+// hostHeap provides an easy way to get best candidate host to assign/steal a shard
+type hostHeap struct {
+	hosts                 []*hostShards
+	rackToHostsMap        map[string]map[*hostShards]struct{}
+	targetLoad2           map[string]int
```
robskillington: What is targetLoad2 and how does it differ from targetLoad? =]

placement/interface.go
```diff
@@ -0,0 +1,104 @@
...
+	Address string
+}
+
+// Service handles the placement related operations for registered services
+// all write or update operations will persist the generated snapshot before returning success
+type Service interface {
+	BuildInitialPlacement(service string, hosts []string, shardLen int) error
+	AddReplica(service string) error
+	AddHost(service string, host string) error
+	AddHostFromPool(service string, pool string) error
```
robskillington: I think this is specific to our use case, let's perhaps add this elsewhere.

placement/interface.go
```diff
@@ -0,0 +1,104 @@
...
+	RemoveHost(service string, host string) error
+	ReplaceHost(service string, leavingHostName string, addingHostName string) error
+	ReplaceHostFromPool(service string, leavingHostName string, pool string) error
```
robskillington: Same here.

placement/interface.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Snapshot gets the persisted snapshot for service
+	Snapshot(service string) (Snapshot, error)
+}
+
+// HostInventory provides inventory information
+type HostInventory interface {
```
robskillington: Instead of having a `HostInventory` perhaps it is better to simply add all this metadata to the `Host` interface?

placement/interface.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// HostShards represents a host and its assigned shards
+type HostShards interface {
+	Host() Host
+	Shards() []uint32
+}
+
+// Host contains the information needed for placement
+type Host struct {
```
robskillington: Might be best to make this an interface so implementers can lazily look up properties on a `Host`

placement/algo/placement.go
```diff
@@ -0,0 +1,266 @@
...
+)
+
+var (
+	errShardsWithDifferentReplicas = errors.New("invalid placement, found shards with different replicas")
+)
+
+// placementSnapshot implements placement.Snapshot
+type placementSnapshot struct {
+	hostShards   []*hostShards
+	shardsLen    int
```
robskillington: Can't this be derived from `len(uniqueShards)`?  Perhaps best to derive on read, especially since this is stored data.

placement/algo/algo.go
cw9: yeah I was doing that because the heap examples seems to be doing that, but then the result needs to be nil-checked then cast, so I changed it to always do a hh.Len()>0 check before Pop(), not sure if it's still necessary to do nil check here

placement/algo/algo.go
cw9: lol targetLoad2 is targetLoad, will rename it

placement/interface.go
cw9: hmm, there is really not much to be lazy-loaded for a Host, I kinda need the address and rack information always, I can make it an interface, but don't think there is much possibility or value for having a different struct/impl of this interface so I turned it into a struct.

placement/algo/placement.go
cw9: yep, will do

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+func (a rackAwarePlacementAlgorithm) AddReplica(ps placement.Snapshot) (placement.Snapshot, error) {
+	ph := newReplicaPlacementHelper(ps, ps.Replicas()+1)
+	if err := ph.placeShards(ph.uniqueShards, nil); err != nil {
+		return nil, err
+	}
+	return ph.generatePlacement(), nil
+}
+
+func (a rackAwarePlacementAlgorithm) RemoveHost(ps placement.Snapshot, leavingHost placement.Host) (placement.Snapshot, error) {
+	if a.isHostAbsent(ps, leavingHost) {
```
xichen2020: maybe this check should be moved inside the `RemoveHostPlacementHelper`?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	var addingHost *hostShards
+	for _, phs := range s.HostShards() {
+		h := newHostShards(phs)
+		if phs.Host().Address == host.Address {
+			addingHost = h
+		}
+		hosts = append(hosts, h)
+	}
+
+	if addingHost == nil {
```
xichen2020: If `addingHost` is not nil, it means the existing placement already has the host we are trying to add in it, so should it be an error?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	for host := range ph.shardToHostMap[shard] {
+		if host.hostRack() == toRack {
+			return false
+		}
+	}
+	return true
+}
+
+func (ph placementHelper) canAssignHost(shard uint32, from, to *hostShards) bool {
+	if to == nil {
```
xichen2020: Can `to` ever be nil? I'd prefer to get rid of the nil checks as much as possible to simplify the logic here.

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	// try to steal shards from the most loaded hosts until the adding host reaches target load
+outer:
+	for len(addingHostShards.shardsSet) < targetLoad {
+		if ph.hostHeap.Len() == 0 {
+			return nil, errCouldNotReachTargetLoad
+		}
+		tryHost := heap.Pop(ph.hostHeap).(*hostShards)
+		for shard := range tryHost.shardsSet {
+			if ph.canAssignHost(shard, tryHost, addingHostShards) {
+				ph.assignShardToHost(shard, addingHostShards)
```
xichen2020: Perhaps we can have a `moveShard(shard, fromHost, toHost)` and call it here instead?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+				ph.removeShardFromHost(shard, tryHost)
+				heap.Push(ph.hostHeap, tryHost)
+				continue outer
```
xichen2020: nit: can just break and get rid of the outer label

placement/algo/algo.go
cw9: good catch, this is particularly for a case in replaceHost, the last step of replace host is to bring the partially loaded addingHost to target load, the newAddHostPlacementHelper could probably be called "bring-a-host-to-target-load-placementhelper"

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+			}
+		}
+	}
+
+	return ph.generatePlacement(), nil
+}
+
+func (a rackAwarePlacementAlgorithm) ReplaceHost(ps placement.Snapshot, leavingHost, addingHost placement.Host) (placement.Snapshot, error) {
+	if a.isHostAbsent(ps, leavingHost) {
```
xichen2020: Yeah seems better to move this inside `newRemoveHostPlacementHelper` so we don't have to check this in multiple places

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+		return nil, errHostAbsent
+	}
+	addingHostShards := newEmptyHostShardsFromHost(addingHost)
+	ph, leavingHostShards := newRemoveHostPlacementHelper(ps, leavingHost)
+
+	var shardsUnassigned []uint32
+	// move shards from leaving host to adding host
+	for _, shard := range leavingHostShards.Shards() {
+		if ph.canAssignHost(shard, leavingHostShards, addingHostShards) {
+			ph.removeShardFromHost(shard, leavingHostShards)
```
xichen2020: I wonder if we could call `MoveShard` here too

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+	// if there are shards that can not be moved to adding host
+	// distribute them to the cluster
+	if err := ph.placeShards(shardsUnassigned, leavingHostShards); err != nil {
+		return nil, err
+	}
+
+	cl := ph.generatePlacement()
+	// add the adding host to the cluster and bring its load up to target load
+	cl.hostShards = append(cl.hostShards, addingHostShards)
```
xichen2020: can we add `addingHostShards` before generating the placement? seems weird that we need to do this as a separate step.

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	return a.AddHost(cl, addingHost)
+}
+
+func (a rackAwarePlacementAlgorithm) isHostAbsent(ps placement.Snapshot, h placement.Host) bool {
+	for _, hs := range ps.HostShards() {
```
xichen2020: Perhaps it'd be better to define a `HasHost` method on the `Snapshot` interface, which internally uses a map to look up a host rather than looping through a list.

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+		}
+		for _, host := range tried {
+			heap.Push(ph.hostHeap, host)
+		}
+		steps = append(steps, parallel)
+	}
+	return steps
+}
+
+type placementHelper struct {
```
xichen2020: Might be structurally more clear to move the helpers into a separate `helper.go` file

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+type shardAwareDeploymentPlanner struct {
+}
+
+func newShardAwareDeploymentPlanner() placement.DeploymentPlanner {
+	return shardAwareDeploymentPlanner{}
+}
+
+func (dp shardAwareDeploymentPlanner) DeploymentSteps(ps placement.Snapshot) [][]placement.HostShards {
+	ph := newReplicaPlacementHelper(ps, ps.Replicas())
+	var steps [][]placement.HostShards
```
xichen2020: nit: might be a bit simpler to do
```
hostHeap := ph.hostHeap
for hostheap.Len() > 0 {
  ...
  var tried HostHeap
  heap.Init(&tried)
  for hostHeap.Len() > 0 {
    ...
    if !h.isSharingShard(*tryHost) {
      ...
    } else {
      tried.Push(tried, tryHost)
    }
  }
  hostHeap = tried
  steps = append(steps, parallel)
}
return steps

File: placement/algo/placement.go [hunk 30]
```diff
@@ -0,0 +1,266 @@
...
+
+	"github.com/m3db/m3cluster/placement"
+)
+
+var (
+	errShardsWithDifferentReplicas = errors.New("invalid placement, found shards with different replicas")
+)
+
+// placementSnapshot implements placement.Snapshot
+type placementSnapshot struct {
```
xichen2020: Can we make this an interface to facilitate testing outside the package and force the clients to interact with it using interface methods instead of directly accessing the internal fields?

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+	shardToHostMap map[uint32]map[*hostShards]struct{}
+	rackToHostsMap map[string]map[*hostShards]struct{}
+	rf             int
+	uniqueShards   []uint32
+	hostShards     []*hostShards
+}
+
+func newInitPlacementHelper(hosts []placement.Host, ids []uint32) *placementHelper {
+	emptyPlacement := newEmptyPlacement(hosts, ids)
+	return newPlaceShardingHelper(emptyPlacement, emptyPlacement.rf+1, true)
```
xichen2020: Can we try to use interface methods as opposed to directly accessing internal fields like `rf` as much as possible? Otherwise it's hard to comprehend how these structs interact with each other.

placement/algo/algo.go
```diff
@@ -0,0 +1,470 @@
...
+type placementHelper struct {
+	hostHeap       *hostHeap
+	shardToHostMap map[uint32]map[*hostShards]struct{}
+	rackToHostsMap map[string]map[*hostShards]struct{}
+	rf             int
+	uniqueShards   []uint32
+	hostShards     []*hostShards
+}
+
+func newInitPlacementHelper(hosts []placement.Host, ids []uint32) *placementHelper {
```
xichen2020: Would it make sense to return an interface instead of a struct pointer? would make it easier to mock out the interface outside the package

placement/algo/algo.go
cw9: in current usage, no, I'll remove this check

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+}
+
+func newReplicaPlacementHelper(s placement.Snapshot, targetRF int) *placementHelper {
+	ps := newPlacementFromGenericSnapshot(s)
+	return newPlaceShardingHelper(ps, targetRF, true)
+}
+
+func newAddHostShardsPlacementHelper(s placement.Snapshot, hs *hostShards) (*placementHelper, *hostShards, error) {
+	var hss []*hostShards
+	for _, phs := range s.HostShards() {
```
xichen2020: Yeah so if we have a `HasHost` method on `Snapshot` interface this could be simplified

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
cw9: done

placement/algo/placement.go
cw9: hmm, not sure if I get your point, but this struct is not supposed to be used outside the package and it's implementing the Snapshot interface

placement/algo/algo.go
cw9: will do

placement/algo/algo.go
cw9: I was not planning to expose these internal structs like placementHelper to outside the package, for example this helper is only helping this implementation of algorithm here, another algorithm probably does not even need such a helper.

There are other structs like placementSnapshot that implements placement.Snapshot interface but again the struct was kept internal for this implementation

placement/interface.go
cw9: done

placement/interface.go
cw9: done

placement/interface.go
cw9: if we are removing the pool related methods from the Service interface, maybe we should remove the Dump(pool string) method here too? Also I don't want to merge this with the Host, Host is just a struct of 2 pieces of data (rack and address) it does not need to know where to retrieve the rack information for a host address?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		ph.rackToHostsMap[h.hostRack()][h] = struct{}{}
+		for shard := range h.shardsSet {
+			ph.assignShardToHost(shard, h)
+		}
+	}
+}
+
+func (ph *placementHelper) buildHostHeap(hostCapacityAscending bool) {
+	overSizedRack := 0
```
xichen2020: Can we move the load calculation logic to a separate `getTargetLoad` method?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+
+	targetLoad := make(map[string]int)
+	for _, host := range ph.hostShards {
+		rackSize := rackSizeMap[host.hostRack()]
+		if float64(rackSize)/float64(totalHostNumber) >= 1.0/float64(ph.rf) {
+			// if the host is on a over-sized rack, the target load is topped at shardLen / rackSize
+			targetLoad[host.hostAddress()] = int(math.Ceil(float64(ph.getShardLen()) / float64(rackSize)))
+		} else {
+			// if the host is on a normal rack, get the target load with aware of other over-sized rack
+			targetLoad[host.hostAddress()] = ph.getShardLen() * (ph.rf - overSizedRack) / (totalHostNumber - overSizedHosts)
```
xichen2020: Just curious, does `targetLoad` need to be an int? 

Also we probably want to do a `ceil` here on the off chance that there are no oversized racks and we have more hosts than `rf * number of shards`.

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+
+func (ph placementHelper) getHostLen() int {
+	return len(ph.hostShards)
+}
+
+func (ph placementHelper) getShardLen() int {
+	return len(ph.uniqueShards)
+}
+
+func (ph placementHelper) getAvgLoad() int {
```
xichen2020: hm..looks like this is only used in the tests. Can we move it to the test file instead and make it a function instead of a method then?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+	totalLoad := ph.rf * ph.getShardLen()
+	numberOfHosts := ph.getHostLen()
+	return totalLoad / numberOfHosts
+}
+
+func (ph placementHelper) getShardSet(shards []uint32) map[uint32]struct{} {
```
xichen2020: This doesn't seem  to depend on `ph`, make it a function instead of a method perhaps?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+	shardSet := make(map[uint32]struct{}, len(shards))
+	for _, shard := range shards {
+		shardSet[shard] = struct{}{}
+	}
+	return shardSet
+}
+
+func (ph placementHelper) canAssignRack(shard uint32, from *hostShards, toRack string) bool {
+	if from != nil {
```
xichen2020: can `from` be nil?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+		return false
+	}
+	if _, exist := to.shardsSet[shard]; exist {
+		return false
+	}
+	return ph.canAssignRack(shard, from, to.hostRack())
+}
+
+func (ph placementHelper) assignShardToHost(shard uint32, to *hostShards) {
+	to.addShard(shard)
```
xichen2020: Do we need to update `ph.hostShrads` too?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+
+	if _, exist := ph.shardToHostMap[shard]; !exist {
+		ph.shardToHostMap[shard] = make(map[*hostShards]struct{})
+	}
+	ph.shardToHostMap[shard][to] = struct{}{}
+}
+
+func (ph placementHelper) removeShardFromHost(shard uint32, from *hostShards) {
+	from.removeShard(shard)
```
xichen2020: Do we need to update `ph.hostShrads` too?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+func (ph placementHelper) placeToRacksOtherThanOrigin(shards map[uint32]struct{}, from *hostShards) {
+	var triedHosts []*hostShards
+	var sameRack []*hostShards
+outer:
+	for shard := range shards {
+		triedHosts = triedHosts[:0]
+		for ph.hostHeap.Len() > 0 {
+			tryHost := heap.Pop(ph.hostHeap).(*hostShards)
+			// do not place to same rack for now
+			if from != nil && tryHost.hostRack() == from.hostRack() {
```
xichen2020: Seems this is trying to find out which hosts are on the same rack as `from`, if that's the case, can we do that outside the for loop? Doing it inside the for loop pollutes the placement logic, making it harder to understand.

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+				sameRack = append(sameRack, tryHost)
+				continue
+			}
+			triedHosts = append(triedHosts, tryHost)
+			if ph.hostHeap.getTargetLoadForHost(tryHost.hostAddress())-len(tryHost.shardsSet) <= 0 {
```
xichen2020: Seems like this assumes the heap is a min heap? Where is that assumption validated?

placement/algo/algo.go
```diff
@@ -0,0 +1,482 @@
...
+}
+
+func (hh hostHeap) Less(i, j int) bool {
+	hostI := hh.hosts[i]
+	hostJ := hh.hosts[j]
+	leftLoadOnI := hh.getTargetLoadForHost(hostI.hostAddress()) - len(hostI.shardsSet)
+	leftLoadOnJ := hh.getTargetLoadForHost(hostJ.hostAddress()) - len(hostJ.shardsSet)
+	// if both host has tokens to be filled, prefer the one on a bigger rack
+	// since it tends to be more picky in accepting shards
+	if leftLoadOnI > 0 && leftLoadOnJ > 0 {
```
xichen2020: Just curious, what's the rationale behind this? Do we want to always pick a host on the bigger rack regardless of the comparator of the heap?

placement/algo/algo.go
cw9: sure

placement/algo/algo.go
cw9: sure

placement/algo/placement.go
xichen2020: yeah so I feel like defining an interface even for internal structs narrows the access surface and hides implementation details so you can swap things out underneath quite easily. Right now we are accessing the internal fields of the structs fairly liberally so it's fairly difficult to make sense of the coupling between `placementSnapshot` and its users.

placement/algo/algo.go
cw9: sure

placement/algo/algo.go
cw9: sure

placement/algo/algo.go
cw9: yep, when adding a full replica, the from will be nil

placement/algo/algo.go
cw9: nope, the to is not necessarily in the cluster yet

placement/algo/algo.go
cw9: same reason as above

placement/algo/algo.go
cw9: yeah will refactor this loop a little

placement/algo/algo.go
cw9: yes, first of all, host will not go out of their target load, I only compare rack when both hosts have room left.
I was only comparing on token left on each host, but would end up with unbalanced distribution sometimes, for example:
replica 1: h1r1: [1], h2r1: [2], h3r2: [3], h4r3:[4]
now we add replica 2, 
if the adding order is 3,4,1,2. because we are looking at the potential capacity on each host, so all 4 hosts are at target - load = 2 - 1 = 1
so shard 3 might end up in h4, shard 4 might end up in h3, then when it comes to 1 and 2, they can only be added to h3 and h4 because of the rack check, thus an unbalanced distribution. This is what I meant about bigger rack being picky in accepting shards

placement/algo/algo.go
cw9: yeah there is a parameter when building the placementHelper, check newPlaceShardingHelper(ps placementSnapshot, targetRF int, hostCapacityAscending bool)

placement/algo/placement.go
cw9: make sense, will do.

placement/algo/helper.go
```diff
@@ -0,0 +1,373 @@
...
+import (
+	"container/heap"
+	"math"
+
+	"github.com/m3db/m3cluster/placement"
+)
+
+// PlacementHelper helps the algorithm to place shards
+type PlacementHelper interface {
+	PlaceShards(shards []uint32, from placement.HostShards) error
```
xichen2020: The interface methods need to be documented

placement/algo/helper.go
```diff
@@ -0,0 +1,373 @@
...
+func (ph *placementHelper) buildTargetLoadMap(rackSizeMap map[string]int, overSizedRackLen, overSizedHostLen int) map[string]int {
+	targetLoad := make(map[string]int)
+	for _, host := range ph.hostShards {
+		rackSize := rackSizeMap[host.Host().Rack()]
+		if float64(rackSize)/float64(ph.getHostLen()) >= 1.0/float64(ph.rf) {
+			// if the host is on a over-sized rack, the target load is topped at shardLen / rackSize
+			targetLoad[host.Host().ID()] = int(math.Ceil(float64(ph.getShardLen()) / float64(rackSize)))
+		} else {
+			// if the host is on a normal rack, get the target load with aware of other over-sized rack
+			targetLoad[host.Host().ID()] = ph.getShardLen() * (ph.rf - overSizedRackLen) / (ph.getHostLen() - overSizedHostLen)
```
xichen2020: what happens if there are more hosts than `rf * number of shards`? Should we do a `math.Ceil` here?

placement/algo/helper.go
```diff
@@ -0,0 +1,373 @@
...
+	if _, exist := ph.shardToHostMap[shard]; !exist {
+		ph.shardToHostMap[shard] = make(map[placement.HostShards]struct{})
+	}
+	ph.shardToHostMap[shard][to] = struct{}{}
+}
+
+func (ph placementHelper) removeShardFromHost(shard uint32, from placement.HostShards) {
+	from.RemoveShard(shard)
+
+	delete(ph.shardToHostMap[shard], from)
```
xichen2020: Do we need to update `ph.hostShards` here?

placement/interface.go
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
xichen2020: nit: just call this file `placement.go`

placement/placement.go
```diff
@@ -0,0 +1,300 @@
...
+		actualTotal += hs.ShardsLen()
+	}
+
+	if expectedTotal != actualTotal {
+		return false
+	}
+	return true
+}
+
+func getUniqueIDSet(ids []uint32) map[uint32]struct{} {
```
xichen2020: This seems to be the same as `convertToShardSet`?

placement/placement.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil, err
+	}
+	return ps, nil
+}
+
+func (ps snapshot) MarshalJSON() ([]byte, error) {
+	return json.Marshal(placementSnapshotToJSON(ps))
+}
+
+func placementSnapshotToJSON(ps snapshot) hostShardsJSONs {
```
xichen2020: Why is this not a method of `snapshot`?

placement/placement.go
```diff
@@ -0,0 +1,300 @@
...
+	hsjs := make(hostShardsJSONs, ps.HostsLen())
+	for i, hs := range ps.hostShards {
+		hsjs[i] = hostShardsToJSON(hs)
+	}
+	sort.Sort(hsjs)
+	return hsjs
+}
+
+func hostShardsToJSON(hs HostShards) hostShardsJSON {
```
xichen2020: can we make this a method of `HostShards`?

placement/placement.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	return hsj[i].Rack < hsj[j].Rack
+}
+
+func (hsj hostShardsJSONs) Swap(i, j int) {
+	hsj[i], hsj[j] = hsj[j], hsj[i]
+}
+
+type hostShardsJSON struct {
+	Address string
```
xichen2020: Perhaps change `Address` to `ID` to be consistent?

placement/placement.go
cw9: you mean the struct of the interface? I kinda don't want to expose this in the interface

placement/placement.go
cw9: will change

placement/placement.go
cw9: yeah did not realize this, but convertToShardSet is in a different package than this one, I guess it's ok to have them both other than shared between packages?

placement/interface.go
cw9: hmm... I already have a placement.go that provides the basic structs for some of the interfaces

placement/algo/helper.go
cw9: nope, the updated hostShard can be in or out of the ph.hostShards, assign or remove a shard does not change that.

placement/placement.go
xichen2020: hm.. I don't like having two methods that are exactly the same. Can we put the implementation somewhere that can be called from different packages?

placement/placement.go
xichen2020: In that case can we do something like `newHostShardsJSON(hs HostShards)`?

placement/algo/helper.go
cw9: will do

placement/algo/helper.go
cw9: Ceil is not necessary, discussed offline

placement/placement.go
cw9: cool, I'll expose it from the placement package

placement/placement.go
cw9: will do

placement/placement.go
cw9: sure

PR #3, opened 2016-08-10
add placement option to loose rack check during host replacement

cw9:
@xichen2020 @robskillington

### DIFF
--- a/placement/algo/algo.go
+++ b/placement/algo/algo.go
@@ -44,7 +44,7 @@ func NewRackAwarePlacementAlgorithm(opt placement.Options) placement.Algorithm {
 }
 
 func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Host, shards []uint32) (placement.Snapshot, error) {
-	ph := newInitPlacementHelper(hosts, shards)
+	ph := newInitPlacementHelper(a.options, hosts, shards)
 
 	if err := ph.PlaceShards(shards, nil); err != nil {
 		return nil, err
@@ -53,7 +53,7 @@ func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Hos
 }
 
 func (a rackAwarePlacementAlgorithm) AddReplica(ps placement.Snapshot) (placement.Snapshot, error) {
-	ph := newPlacementHelperWithTargetRF(ps, ps.Replicas()+1)
+	ph := newPlacementHelperWithTargetRF(a.options, ps, ps.Replicas()+1)
 	if err := ph.PlaceShards(ps.Shards(), nil); err != nil {
 		return nil, err
 	}
@@ -64,7 +64,7 @@ func (a rackAwarePlacementAlgorithm) RemoveHost(ps placement.Snapshot, leavingHo
 	var ph PlacementHelper
 	var leavingHostShards placement.HostShards
 	var err error
-	if ph, leavingHostShards, err = newRemoveHostPlacementHelper(ps, leavingHost); err != nil {
+	if ph, leavingHostShards, err = newRemoveHostPlacementHelper(a.options, ps, leavingHost); err != nil {
 		return nil, err
 	}
 	// place the shards from the leaving host to the rest of the cluster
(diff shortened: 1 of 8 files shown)


placement/algo/algo.go
```diff
@@ -74,22 +75,31 @@ func (a rackAwarePlacementAlgorithm) RemoveHost(ps placement.Snapshot, leavingHo
...
 	if ph, leavingHostShards, err = newRemoveHostPlacementHelper(ps, leavingHost); err != nil {
 		return nil, err
 	}
-	addingHostShards := placement.NewEmptyHostShardsFromHost(addingHost)
+
 	var shardsUnassigned []uint32
 	// move shards from leaving host to adding host
 	for _, shard := range leavingHostShards.Shards() {
-		if moved := ph.MoveShard(shard, leavingHostShards, addingHostShards); !moved {
+		if moved := ph.MoveShard(shard, leavingHostShards, addingHostShards, !a.options.LooseRackCheck()); !moved {
```
xichen2020: Instead of passing the boolean flag to the `MoveShard` method, I wonder if it makes more sense to pass the options to the PlacementHelper constructor so its `MoveShard` method behaves differently depending on the options. This way we don't have to expose it in the `PlacementHelper` interface, and if we need to add/relax more constraints in the future, it's easy to do so.

placement/algo/helper.go
```diff
@@ -227,7 +229,7 @@ func (ph placementHelper) getShardLen() int {
 	return len(ph.uniqueShards)
 }
 
-func (ph placementHelper) canAssignRack(shard uint32, from placement.HostShards, toRack string) bool {
+func (ph placementHelper) CanMoveShardToRack(shard uint32, from placement.HostShards, toRack string) bool {
```
xichen2020: Perhaps call this `HasNoRackConflict`? We might decide to move `shard` to `toRack` even if this function returns false if the rack constraint is loosened.

placement/algo/helper.go
```diff
@@ -241,11 +243,14 @@ func (ph placementHelper) canAssignRack(shard uint32, from placement.HostShards,
 	return true
 }
 
-func (ph placementHelper) canAssignHost(shard uint32, from, to placement.HostShards) bool {
+func (ph placementHelper) canAssignHost(shard uint32, from, to placement.HostShards, rackCheck bool) bool {
```
xichen2020: yeah if the placementHelper stores the placement option inside, it doesn't need the `rackCheck` flag to be passed in.

placement/service/placementservice.go
```diff
@@ -219,13 +209,73 @@ func findBestHost(ps placementService, p placement.Snapshot, candidateHosts []pl
...
+	if hs, exist := candidateRackHostMap[leaving.Host().Rack()]; exist {
+		return hs[0], nil
+	}
+
+	return findHostWithRackConflictCheck(s, candidateRackHostMap, leaving, looseRackCheck)
+}
+
+func findHostWithRackConflictCheck(p placement.Snapshot, rackHostMap map[string][]placement.Host, leaving placement.HostShards, looseRackCheck bool) (placement.Host, error) {
+	// otherwise sort the candidate hosts by the number of conflicts
+	ph := algo.NewPlacementHelper(p)
```
xichen2020: If we are constructing a `PlacementHelper` inside the function, maybe this function should just be a method of `PlacementHelper`?

placement/algo/algo.go
cw9: sure will do

placement/algo/helper.go
cw9: Will rename, but in that case I will not do this check if the constraint is loosened

placement/algo/helper.go
cw9: sure

placement/service/placementservice.go
cw9: I just need to use the HasNoRackConflict() check from the ph here, so I feel it's fine to leave this function here

placement/algo/helper.go
```diff
@@ -38,9 +38,9 @@ type PlacementHelper interface {
 	// MoveOneShard moves one shard between 2 hosts
 	MoveOneShard(from, to placement.HostShards) bool
 	// MoveShard moves a particular shard between 2 hosts
-	MoveShard(shard uint32, from, to placement.HostShards, rackCheck bool) bool
-	// CanMoveShardToRack checks if it's valid to move a shard to target rack
-	CanMoveShardToRack(shard uint32, from placement.HostShards, toRack string) bool
+	MoveShard(shard uint32, from, to placement.HostShards) bool
+	// HasNoRackConflict checks if it's valid to move a shard to target rack
```
xichen2020: `checks if the rack constraint is violated if the given shard is moved to the target rack`

placement/algo/helper.go
```diff
@@ -243,14 +264,14 @@ func (ph placementHelper) CanMoveShardToRack(shard uint32, from placement.HostSh
...
 }
 
-func (ph placementHelper) canAssignHost(shard uint32, from, to placement.HostShards, rackCheck bool) bool {
+func (ph placementHelper) canAssignHost(shard uint32, from, to placement.HostShards) bool {
 	if to.ContainsShard(shard) {
 		return false
 	}
-	if rackCheck {
-		return ph.CanMoveShardToRack(shard, from, to.Host().Rack())
+	if ph.options.LooseRackCheck() {
```
xichen2020: nit: can be `ph.options.LooseRackCheck() || ph.HasNoRackConflict(shard, from, to.Host().Rack())`

placement/algo/helper.go
xichen2020: Yeah I meant the new function name would be more consistent with its purpose because it seemed a bit weird if `CanMoveShardToRack` returns true and we assign the shard to the rack anyway.

xichen2020:
LGTM

PR #4, opened 2016-08-10
Marshal placement snapshot as maps instead of arrays

xichen2020:
cc @cw9 @robskillington

### DIFF
--- a/placement/placement.go
+++ b/placement/placement.go
@@ -125,13 +125,12 @@ func (ps snapshot) MarshalJSON() ([]byte, error) {
 	return json.Marshal(ps.placementSnapshotToJSON())
 }
 
-func (ps snapshot) placementSnapshotToJSON() hostShardsJSONs {
-	hsjs := make(hostShardsJSONs, ps.HostsLen())
-	for i, hs := range ps.hostShards {
-		hsjs[i] = newHostShardsJSON(hs)
+func (ps snapshot) placementSnapshotToJSON() map[string]hostShardsJSON {
+	m := make(map[string]hostShardsJSON, ps.HostsLen())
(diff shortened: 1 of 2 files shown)


placement/placement.go
```diff
@@ -156,28 +155,30 @@ func (su sortableUInt32) Swap(i, j int) {
...
 }
 
-func convertJSONtoSnapshot(hsjs hostShardsJSONs) (snapshot, error) {
+func convertJSONtoSnapshot(m map[string]hostShardsJSON) (snapshot, error) {
 	var err error
-	hss := make([]HostShards, len(hsjs))
+	hss := make([]HostShards, len(m))
 	shardsReplicaMap := make(map[uint32]int)
-	for i, hsj := range hsjs {
+	i := 0
```
cw9: is this i used?

placement/placement_test.go
```diff
@@ -160,23 +160,23 @@ func TestValidate(t *testing.T) {
...
-	]`
+	invalidJSON := `{
+		:{"ID":123,"Rack":"r1","Shards":[0,7,11]}
+	}`
 	data := []byte(invalidJSON)
 	ps, err := NewPlacementFromJSON(data)
 	assert.Nil(t, ps)
 	assert.Error(t, err)
 
-	validJSON := `[
```
cw9: the order is no longer guaranteed is it?

placement/placement.go
xichen2020: yeah in the for loop

placement/placement_test.go
xichen2020: no it's not, but apparently the diff is smaller with maps if we change the placements

cw9:
LGTM

PR #7, opened 2016-08-17
Flesh out KV support

mmihic:


Makefile
```diff
@@ -42,6 +50,25 @@ test-ci-unit: test-internal
 	@which goveralls > /dev/null || go get -u -f github.com/mattn/goveralls
 	goveralls -coverprofile=$(coverfile) -service=travis-ci || echo -e "\x1b[31mCoveralls failed\x1b[m"
 
+install-vendor: .gitmodules
```
xichen2020: Should this be glide based now that we've migrated over to glide?

mmihic: We still need submodules for uber-licence

kv/mock.go
```diff
@@ -0,0 +1,125 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package kv
+
+import (
+	"github.com/golang/protobuf/proto"
+	"sync"
```
xichen2020: Reorder the import path perhaps?

Makefile
mmihic: We could move this out of vendor though so it's clear this isn't a golang dep

vendor/github.com/uber/uber-licence/license-fixer.js
```diff
@@ -0,0 +1,179 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: Hm, this file is showing up as an add?  Shouldn't this be a submodule?

mmihic: Hm, it was added as a submodule.  Let me try to re-apply in some way.

Makefile
xichen2020: ah yeah, forgot about the node module.

I'm fine either way. Someone can probably spend a day rewriting the license module in go so we don't end up having a node module in go repos :).

PR #8, opened 2016-08-21
Change Set support

mmihic:


changeset/manager.go
```diff
@@ -0,0 +1,415 @@
...
+		panic("opts.KV must be set")
+	}
+
+	if opts.GetConfigType() == nil {
+		panic("opts.ConfigType must be set")
+	}
+
+	if opts.GetChangesType() == nil {
+		panic("opts.ChangesType must be set")
+	}
```
robskillington: For opts checking the pattern we've been using is implementing a `Validate() error` method on the options itself if you wanted to use the same pattern.  Also perhaps better to return a `(Manager, error)` and then `if err := opts.Validate(); err != nil { return nil, err }` than panic?

mmihic: panic on programming error is golang-style, but sure I can switch to return an error instead.

changeset/manager.go
```diff
@@ -0,0 +1,415 @@
...
+			return err
+		}
+
+		changeBytes, err := proto.Marshal(changes)
+		if err != nil {
+			return err
+		}
+
+		// ...and update the stored changes
+		changeset.Changes = changeBytes
```
robskillington: So changeset has a serialized version of itself?  Interesting.  I guess `.Changes` always keeps the previous version of the changes, is this for rollback purposes?

mmihic: No, ``ChangeSet`` is a wrapper around the actual changes, adding in the state of those changes.  ``.Changes`` contains the actual user-defined changes (in serialized form).

robskillington: I see, so for a given `configType`, in order to make multiple changes, the underlying `changesType` must allow for the caller to inspect the previous changes inside the saved `changesType` and mutate it such that both all that came before it and its own changes are captured in it by the mutations it makes itself to instance of the `changesType` passed to it.  

Ok I get it now.  This means we won't natively be able to capture the length of modifications made nor stringify each one individually.  I think this makes sense but because we don't explicitly store them separately it will reduce the capability somewhat of tooling that only has access to a concrete `Manager`.

robskillington: Also is there anyway possibly to know if the `changesType` passed to you is a default value (i.e. you are the first change) or otherwise?

mmihic: The ``Manager`` is intended to be a support structure; tools will use code that understands the concrete types.  I guess we could maintain the changes as a repeated set of serialized messages so you could manipulate them individually, but that seems overkill.

You'd know you are the first change by looking at the ``changesType`` and seeing it doesn't have any changes. Don't see a need for the framework to get involved in that.

changeset/manager.go
```diff
@@ -0,0 +1,432 @@
...
+
+		return err
+	}
+
+	// And mark the changes as ommitted
+	// TODO(mmihic): There is an error case here where the changes could be committed by the
+	// the changeset not updated. This is harmeless since the version number of the config will
+	// have been updated so we'll never return to this changeset again, but it does leave the
+	// changeset with the incorrect state of COMMITTING. We could have a mechanism to fix this,
+	// either automatically or by command
```
robskillington: Could we perhaps just allow this command to be run again?  It seems we only check `== COMMITTED` anyhow, this might work out of the box?

mmihic: No, because at that point the "current" config version will have moved on past the version you are trying to commit.  I guess at that point we could check to see if the changes are marked ``COMMITTED`` and mark them as such if they are not.

robskillington: Oh I see, at this point they've already been committed.  Yeah this seems harmless.  `GetPendingChanges(...)` should return an empty set anyway if it reads the latest config and config version tuple.

changeset/manager_test.go
```diff
@@ -0,0 +1,1052 @@
...
+		// Fetch corresponding changes
+		s.kv.EXPECT().Get("config/_changes/2").Return(changesVal, nil),
+
+		// ...And update
+		s.kv.EXPECT().CheckAndSet("config/_changes/2", 12, gomock.Any()).Return(13, nil),
+	)
+
+	require.NoError(t, s.mgr.Change(addLines("foo", "bar")))
+
+	// NB(mmihic): We only care that the expectations are met
```
robskillington: Don't want to check that config is now "foo\nbar"?

robskillington: Don't want to check that config is now "foo\nbar"?

mmihic: Nah, we've already test that path.

robskillington:
LGTM

changeset/manager.go
```diff
@@ -0,0 +1,432 @@
...
+	if err := opts.Validate(); err != nil {
+		return nil, err
+	}
+
+	logger := opts.GetLogger()
+	if logger == nil {
+		logger = xlog.NullLogger
+	}
+
+	return &manager{
```
xichen2020: Since the member methods are all defined on the value and the manager instance is essentially immutable, could probably return a value here to avoid a heap allocation?

changeset/manager.go
```diff
@@ -0,0 +1,432 @@
...
+		return err
+	}
+
+	var changeset changesetpb.ChangeSet
+	if err := csVal.Unmarshal(&changeset); err != nil {
+		return err
+	}
+
+	// If the change list is already committed, bail out
+	if changeset.State == changesetpb.ChangeSetState_COMMITTED {
```
xichen2020: Say a `Commit` call gets through here and sets the changeset state to committing, but before it bumps up the config version and sets the changeset state to commited, another `Commit` call comes in, sees the changeset state is `ChangeSetState_COMMITTING`, and bumps up the changeset version again. Now the first `Commit` call bumps up the config version, but will fail to set the change set version since it's been updated by the second `Commit` call, and the second `Commit` call will fail when committing the new config since it's been updated by the first `Commit` call.

Do we want to protect ourselves against this case by checking the state against `changesetpb.ChangeSetState_COMMITTING`?

changeset/manager.go
mmihic: Was thinking about this more - and I think I'm going to get rid of the COMMITTED state entirely (along with this second update).  Basically, ChangeSets will either be OPEN (accepting new changes) or CLOSED (commit started, new changes not allowed).  You know that a CLOSED ChangeSet has been committed because the config object has moved past the version on which the ChangeSet was based (i.e. if config is at v72, you know that the ChangeSet for v71 was committed).  

May also add a COMMIT_FAILED state, for when the applyFn passed to commit returns an error.  This is usually fatal (basically, the changes could not properly transform the config, probably due to a bug) and flagging this separately would allow us to have a transaction-safe Rollback operation which could reset failed changesets to an empty state for re-application.

changeset/manager.go
mmihic: These are going to long-lived singletons, so I'm not too concerned about the heap allocation.  But yeah, can do.

changeset/manager.go
mmihic: Good catch.  We still want to allow commits against a COMMITTING ChangeSet to proceed, otherwise transient failures (such as network errors) that occur during Commit() will render the ChangeSet unusable without manual repair. But we can avoid this particular race condition by just not updating the ChangeSet if its state is already COMMITTING.

changeset/manager.go
mmihic: Was thinking about this more - and def don't want to impose a more granular structure on changes.  The storage placement stuff, for example, isn't going to handle changes as discrete chunks - it needs to track them as a global modification set indexed by affected entity (database, cluster, etc).  Forcing denormalization into separate chunks would significantly complicate that.

PR #9, opened 2016-08-22
Add WatchInstances to Service interface

cw9:
@xichen2020 @robskillington @martin-mao @mmihic

### DIFF
--- a/client.go
+++ b/client.go
@@ -22,6 +22,7 @@ package cluster
 
 import (
 	"github.com/m3db/m3cluster/kv"
+	"github.com/m3db/m3x/watch"
 )
 
 // A ServiceInstance is a single instance of a service
@@ -38,6 +39,17 @@ type ServiceInstance interface {
 
 // NewServiceInstance creates a new ServiceInstance
 func NewServiceInstance() ServiceInstance { return new(serviceInstance) }
+
+// A Shard represents a piece of data owned by the service
+type Shard interface {
+	ID() uint32
+}
+
+// InstanceAssignment represents a ServiceInstance and the shards owned by it
+type InstanceAssignment interface {
+	ServiceInstance() ServiceInstance // ServiceInstance returns the ServiceInstance
+	Shards() Shard                    // Shards returns the shards
+}
 
 // Advertisement advertises the availability of a given instance of a service
 type Advertisement interface {
@@ -75,6 +87,9 @@ type Services interface {
 
 	// QueryInstances returns the list of available instances for a given service
 	QueryInstances(service string, opts QueryOptions) ([]ServiceInstance, error)
+
+	// WatchAssignment returns a watch on InstanceAssignment updates for a given service
+	WatchAssignment(service string, opts QueryOptions) (watch.Watch, error)
 }
 
 // Client is the base interface into the cluster management system, providing
--- a/client.go
+++ b/client.go
@@ -25,23 +25,6 @@ import (
 	"github.com/m3db/m3x/watch"
 )
 
-// A Shard represents a piece of data owned by the service
-type Shard interface {
-	ID() uint32 // ID returns the id of the shard
-}
-
-// NewShard returns a new Shard
-func NewShard(s uint32) Shard { return &shard{id: s} }
-
-// Shards is a collection of shards owned by one ServiceInstance
-type Shards interface {
-	Shards() []Shard // Shards returns the shards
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -75,6 +87,9 @@ type Services interface {
 
 	// QueryInstances returns the list of available instances for a given service
 	QueryInstances(service string, opts QueryOptions) ([]ServiceInstance, error)
+
+	// WatchAssignment returns a watch on InstanceAssignment updates for a given service
+	WatchAssignment(service string, opts QueryOptions) (watch.Watch, error)
```
robskillington: I think `WatchInstances` might be better here to match the above method.

client.go
```diff
@@ -39,6 +40,17 @@ type ServiceInstance interface {
...
 
+// A Shard represents a piece of data owned by the service
+type Shard interface {
+	ID() uint32
+}
+
+// InstanceAssignment represents a ServiceInstance and the shards owned by it
+type InstanceAssignment interface {
+	ServiceInstance() ServiceInstance // ServiceInstance returns the ServiceInstance
+	Shards() Shard                    // Shards returns the shards
```
robskillington: Shouldn't this be `[]Shard` as the result type?

client.go
```diff
@@ -39,6 +40,17 @@ type ServiceInstance interface {
 // NewServiceInstance creates a new ServiceInstance
 func NewServiceInstance() ServiceInstance { return new(serviceInstance) }
 
+// A Shard represents a piece of data owned by the service
+type Shard interface {
+	ID() uint32
+}
+
+// InstanceAssignment represents a ServiceInstance and the shards owned by it
+type InstanceAssignment interface {
```
robskillington: Where is this actually returned from?  I don't see a place for this yet?

client.go
```diff
@@ -22,8 +22,25 @@ package cluster
 
 import (
 	"github.com/m3db/m3cluster/kv"
+	"github.com/m3db/m3x/watch"
 )
 
+// A Shard represents a piece of data owned by the service
+type Shard interface {
```
xichen2020: Why is `Shard` in client.go? Can we move it to `shard.go`?

client.go
```diff
@@ -22,8 +22,25 @@ package cluster
...
+	ID() uint32 // ID returns the id of the shard
+}
+
+// NewShard returns a new Shard
+func NewShard(s uint32) Shard { return &shard{id: s} }
```
xichen2020: nit: don't need to return a pointer here. The id is immutable after the shard is created, so can just be a value and avoid a heap allocation.

client.go
```diff
@@ -22,8 +22,25 @@ package cluster
...
+
+// Shards is a collection of shards owned by one ServiceInstance
+type Shards interface {
+	Shards() []Shard // Shards returns the shards
+}
+
+// NewShards returns a new instance of Shards
+func NewShards(ss []Shard) Shards { return &shards{ss: ss} }
```
xichen2020: Return a value here?

client.go
robskillington: Technically it will still be a heap allocation as the interface type is allocated on the heap and will hold a pointer to the struct.  Still better though as it's not a pointer to a pointer so one less mem access when retrieving. Heh.

client.go
```diff
@@ -75,6 +94,9 @@ type Services interface {
 
 	// QueryInstances returns the list of available instances for a given service
 	QueryInstances(service string, opts QueryOptions) ([]ServiceInstance, error)
+
+	// WatchInstances returns a watch on instances updates for a given service
+	WatchInstances(service string, opts QueryOptions) (watch.Watch, error)
```
robskillington: Wondering if we should've named the package `xwatch` like we do with all the other top level `m3x` packages?  cc @mmihic @xichen2020

client.go
cw9: sure

client.go
cw9: we probably should to keep the pattern in sync, will change

client.go
mmihic: It won't be a pointer to a pointer either; the interface will be the ref directly.

client.go
cw9: the package name change will be applied in a follow up diff

xichen2020: +1 on renaming it to xwatch, should keep it consistent

robskillington:
LGTM

PR #11, opened 2016-08-23
Use xwatch instead of watch

cw9:


xichen2020:
Do we need to update glide to include the updated version of m3x?

xichen2020:
LGTM

PR #12, opened 2016-08-24
Add client mock

cw9:
@xichen2020 @robskillington


cw9:
Wondering if client.go should be moved to client/ so the mock does not stay on top level

client.go
```diff
@@ -22,7 +22,7 @@ package cluster
 
 import (
 	"github.com/m3db/m3cluster/kv"
-	"github.com/m3db/m3x/watch"
+	xwatch "github.com/m3db/m3x/watch"
```
xichen2020: Don't think you need this since the package name should already be xwatch

generated/mocks/generate.go
```diff
@@ -19,5 +19,6 @@
 // THE SOFTWARE.
 
 //go:generate sh -c "mockgen -package=kv -destination=$GOPATH/src/$PACKAGE/kv/mock_store.go -source=$GOPATH/src/$PACKAGE/kv/store.go"
+//go:generate sh -c "mockgen -package=cluster -destination=$GOPATH/src/$PACKAGE/mock_client.go -source=$GOPATH/src/$PACKAGE/client.go"
```
xichen2020: Not for this diff, but eventually this should be a build target

xichen2020:
Yeah it makes sense to move `client.go` to its own package

generated/mocks/generate.go
xichen2020: Also probably change the package name to `client` if you move `client.go` to its own package

client.go
cw9: that way the mockgen would fail, it could not find the xwatch package, only works with rename somehow.

client/client.go
```diff
@@ -18,11 +18,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package cluster
+package client
 
 import (
 	"github.com/m3db/m3cluster/kv"
-	"github.com/m3db/m3x/watch"
+	xwatch "github.com/m3db/m3x/watch"
```
xichen2020: can we get rid of the alias here? the package name in m3x/watch is already xwatch

xichen2020:
LGTM

PR #14, opened 2016-08-28
Add ServiceReplication and ServiceSharding

robskillington:
cc @cw9 @xichen2020


shard/shard.go
```diff
@@ -49,3 +53,28 @@ type shards struct {
 
 func (s shards) Shards() []Shard { return s.ss }
 func (s shards) ShardsLen() int  { return len(s.ss) }
+
+// ShardingInfo is a set of information for a service's sharding
+type ShardingInfo interface {
```
cw9: How about remove the Shards info from ServiceInstances, and return the ServiceInstance-Shards data here in the ShardingInfo, this way ServiceInstances can be directly used by non-sharded services? 

And have a WatchShardingInfo function to subscribe to ShardingInfo updates?

services/services.go
```diff
@@ -25,7 +25,40 @@ import (
...
+	// Replicas is the count of replicas
+	Replicas() int
+}
+
+// NewServiceReplication creates a new ServiceReplication
+func NewServiceReplication() ServiceReplication { return new(serviceReplication) }
+
+// ServiceSharding describes the sharding of a service
+type ServiceSharding interface {
+	// Range is the count of shards to shard over
```
cw9: nit: rename Range to Count? // Count is the count... ?

cw9:
LGTM

PR #15, opened 2016-08-28
Add missing setters to Service, ServiceReplication and ServiceSharding

robskillington:
cc @xichen2020 @cw9

### DIFF
--- a/services/services.go
+++ b/services/services.go
@@ -33,8 +33,14 @@ type Service interface {
 	// Replication returns the service replication description or nil if none
 	Replication() ServiceReplication
 
+	// SetReplication sets the service replication description or nil if none
+	SetReplication(r ServiceReplication) Service
+
 	// Sharding returns the service sharding description or nil if none
 	Sharding() ServiceSharding
+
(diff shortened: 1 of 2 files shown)


cw9:
LGTM

mmihic:
They weren't missing - they were intentionally left out.  The objects returned from the cluster API should be immutable.

services/services.go
```diff
@@ -53,6 +65,9 @@ func NewServiceReplication() ServiceReplication { return new(serviceReplication)
 type ServiceSharding interface {
 	// Len is the count of shards to use as the sharding range
 	Len() int
```
mmihic: ``Len`` seems like a misleading terminology for this.  In go-land ``Len`` typically denotes the number of items in a container, in this case its really the cardinality of a numeric space which seems more like a ``Count`` or (to use Google style) ``NumShards``.

robskillington: Have we used a term in storage for this yet?  We should just copy if so.

mmihic: NumShards is used in storage

PR #19, opened 2016-08-31
Adding Subscribe to the KV Store interface

martin-mao:


robskillington:
LGTM

kv/store.go
```diff
@@ -49,11 +50,25 @@ type Value interface {
...
+}
+
 // Store provides access to the configuration store
 type Store interface {
 	// Get retrieves the value for the given key
 	Get(key string) (Value, error)
 
+	// Watch adds a watch for value updates for given key. Initial value
+	// will be available via ValueWatch.Get()
+	Watch(key string) (ValueWatch, error)
```
mmihic: We should call out explicitly whether this will block until the initial value is available.

PR #29, opened 2016-10-23
Consider weights in placement algorithm

cw9:
cc @xichen2020 @robskillington

### DIFF
--- a/placement/algo/algo.go
+++ b/placement/algo/algo.go
@@ -23,6 +23,7 @@ package algo
 import (
 	"container/heap"
 	"errors"
+	"fmt"
 
 	"github.com/m3db/m3cluster/placement"
 )
@@ -44,7 +45,7 @@ func NewRackAwarePlacementAlgorithm(opt placement.Options) placement.Algorithm {
 }
 
 func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Host, shards []uint32) (placement.Snapshot, error) {
-	ph := newInitPlacementHelper(a.options, hosts, shards)
+	ph := newInitHelper(hosts, shards, a.options)
 
 	if err := ph.PlaceShards(shards, nil); err != nil {
 		return nil, err
@@ -53,7 +54,8 @@ func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Hos
 }
 
 func (a rackAwarePlacementAlgorithm) AddReplica(ps placement.Snapshot) (placement.Snapshot, error) {
-	ph := newPlacementHelperWithTargetRF(a.options, ps, ps.Replicas()+1)
+	ps = placement.CopySnapshot(ps)
+	ph := newAddReplicaHelper(ps, a.options)
 	if err := ph.PlaceShards(ps.Shards(), nil); err != nil {
 		return nil, err
 	}
@@ -61,10 +63,9 @@ func (a rackAwarePlacementAlgorithm) AddReplica(ps placement.Snapshot) (placemen
 }
 
 func (a rackAwarePlacementAlgorithm) RemoveHost(ps placement.Snapshot, leavingHost placement.Host) (placement.Snapshot, error) {
-	var ph PlacementHelper
-	var leavingHostShards placement.HostShards
-	var err error
-	if ph, leavingHostShards, err = newRemoveHostPlacementHelper(a.options, ps, leavingHost); err != nil {
+	ps = placement.CopySnapshot(ps)
+	ph, leavingHostShards, err := newRemoveHostHelper(ps, leavingHost, a.options)
+	if err != nil {
 		return nil, err
 	}
 	// place the shards from the leaving host to the rest of the cluster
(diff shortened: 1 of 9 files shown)


robskillington:
Only managed to skim this, will review in full tomorrow 👍

placement/algo/algo.go
```diff
@@ -75,45 +76,61 @@ func (a rackAwarePlacementAlgorithm) RemoveHost(ps placement.Snapshot, leavingHo
...
+	for _, addingHost := range addingHosts {
+		newPS, leavingHostShards, err = removeHostFromPlacement(newPS, addingHost)
+		if err != nil {
+			return nil, err
+		}
+		newPS, err = a.addHostShards(newPS, leavingHostShards)
+		if err != nil {
+			return nil, err
+		}
+	}
```
cw9: yeah so addHostShards() would add a HostShards object to the placement and bring it's capacity to target load, this loop here is for cases when some of the new hosts could not reach it's ideal target load due to, say, rack conflict, when taking shards from the leaving host. This operation here would move some shards from some other loaded hosts to this under loaded host. 

The removeHostFromPlacement() will not remove all the shards from the HostShards object, it will just remove it from the placement so I can call addHostShards() to add it back again.

At this moment, we will only be setting a.options.AllowPartialReplace() = false, so we will never do this shard distribution optimization. But in the future this could help maintaining a evenly distributed cluster

placement/algo/helper.go
```diff
@@ -78,192 +82,216 @@ func (ph *placementHelper) MoveShard(shard uint32, from, to placement.HostShards
...
 
-func newPlacementHelperWithTargetRF(opt placement.Options, s placement.Snapshot, targetRF int) PlacementHelper {
-	return newPlaceShardingHelper(opt, s, targetRF, true)
+func newAddReplicaHelper(ps placement.Snapshot, opt placement.Options) PlacementHelper {
+	return newHelper(ps, ps.Replicas()+1, opt)
 }
 
-func newAddHostShardsPlacementHelper(opt placement.Options, s placement.Snapshot, hs placement.HostShards) PlacementHelper {
-	var hss []placement.HostShards
+func newAddHostShardsPlacementHelper(ps placement.Snapshot, hostShards placement.HostShards, opt placement.Options) PlacementHelper {
```
robskillington: Maybe we should rename this too from `newAddHostShardsPlacementHelper` to just `newAddHostShardsHelper`?

placement/service/placementservice.go
```diff
@@ -338,3 +391,22 @@ func (ps placementService) validateInitHosts(hosts []placement.Host) error {
 	}
 	return nil
 }
+
+type sortableThing struct {
```
robskillington: Perhaps a better name is `sortableValue`?

cw9: yeah will do

placement/types.go
```diff
@@ -36,8 +36,8 @@ type Algorithm interface {
 	// RemoveHost removes a host from the cluster
 	RemoveHost(p Snapshot, h Host) (Snapshot, error)
 
-	// ReplaceHost replace a host with a new host
-	ReplaceHost(p Snapshot, leavingHost, addingHost Host) (Snapshot, error)
+	// ReplaceHost replace a host with new hosts
+	ReplaceHost(p Snapshot, leavingHost Host, addingHost []Host) (Snapshot, error)
```
robskillington: This should probably be `addingHosts` for the parameter?  (just a method signature but probably worth updating)

cw9: yep will do

placement/service/placementservice.go
```diff
@@ -233,39 +222,112 @@ func (ps placementService) findReplaceHost(s placement.Snapshot, candidateHosts
...
+	if totalWeight <= targetWeight {
+		return hosts[:], targetWeight - totalWeight
+	}
+	// totalWeight > targetWeight, there is a combination of hosts to meet targetWeight for sure
+	// we do dp until totalWeight rather than targetWeight here because we need to cover the targetWeight
+	// which is a little bit different than the knapsack problem that goes
+	weights := make([]int, totalWeight+1)
+	combination := make([][]placement.Host, totalWeight+1)
+
+	for _, host := range hosts {
```
xichen2020: Can we add some comments here to explain the DP relationship and that we are preferring assignments that have fewer hosts in the assignment?

placement/service/placementservice.go
```diff
@@ -233,39 +222,112 @@ func (ps placementService) findReplaceHost(s placement.Snapshot, candidateHosts
...
-		return rackHostMap[rackLens[0].rack][0], nil
+	for i := targetWeight; i <= totalWeight; i++ {
+		if weights[i] >= targetWeight {
+			return combination[i], targetWeight - weights[i]
+		}
 	}
-	return nil, errNoValidHost
+
+	// will never reach here
+	return nil, 0
```
xichen2020: If we should never reach here, should we log an error or even panic to catch the bug etc?

placement/algo/algo.go
```diff
@@ -53,18 +54,18 @@ func (a rackAwarePlacementAlgorithm) BuildInitialPlacement(hosts []placement.Hos
 }
 
 func (a rackAwarePlacementAlgorithm) AddReplica(ps placement.Snapshot) (placement.Snapshot, error) {
-	ph := newPlacementHelperWithTargetRF(a.options, ps, ps.Replicas()+1)
+	ps = placement.CopySnapshot(ps)
```
cw9: yeah right, will do

placement/service/placementservice.go
cw9: yeah will do

placement/service/placementservice.go
cw9: will do

placement/service/placementservice.go
```diff
@@ -392,12 +392,12 @@ func (ps placementService) validateInitHosts(hosts []placement.Host) error {
 	return nil
 }
 
-type sortableThing struct {
+type sortableValue struct {
 	thing interface{}
```
cw9: yeah will do

PR #30, opened 2016-10-26
Add DeploymentOptions to DeploymentPlanner

cw9:
cc @xichen2020 @robskillington 

Save some steps in deployment than cutting steps by the caller.


placement/planner/planner.go
```diff
@@ -26,47 +26,57 @@ import (
 	"github.com/m3db/m3cluster/placement"
 )
 
-// shardAwareDeploymentPlanner plans the deployment so that as many hosts can be deployed
-// at the same time without making more than 1 replica of any shard unavailable
-type shardAwareDeploymentPlanner struct {
+const defaultMaxStepSize = 3
```
xichen2020: Perhaps better to move this to `options.go`?

PR #33, opened 2016-11-01
Sort hosts by id before knapsack

cw9:
Make the replace function deterministic given the same set of hosts

### DIFF
--- a/placement/placement.go
+++ b/placement/placement.go
@@ -336,17 +336,17 @@ func (h host) String() string {
 	return fmt.Sprintf("[id:%s, rack:%s, zone:%s, weight:%v]", h.id, h.rack, h.zone, h.weight)
 }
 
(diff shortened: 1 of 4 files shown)


cw9:
@xichen2020 @robskillington

placement/placement.go
```diff
@@ -334,3 +335,18 @@ func (h host) Weight() uint32 {
 func (h host) String() string {
 	return fmt.Sprintf("[id:%s, rack:%s, zone:%s, weight:%v]", h.id, h.rack, h.zone, h.weight)
 }
+
+// SortHostByID sorts Hosts by ID
+type SortHostByID []Host
```
xichen2020: `ByIDAscending` might be a better name?

cw9: sure, will fix

PR #34, opened 2016-11-08
Service Discovery Service interface change

cw9:
@robskillington @xichen2020 @martin-mao

### DIFF
--- a/services/services.go
+++ b/services/services.go
@@ -21,13 +21,9 @@
 package services
 
 import (
-	"errors"
-
 	"github.com/m3db/m3cluster/shard"
 	xwatch "github.com/m3db/m3x/watch"
 )
-
-var errInstanceNotFound = errors.New("instance not found")
 
 // Service describes the metadata and instances of a service
 type Service interface {
@@ -53,8 +49,29 @@ type Service interface {
 	SetSharding(s ServiceSharding) Service
 }
 
-// NewService creates a new Service
-func NewService() Service { return new(service) }
+// PlacementSnapshot describes how shards are placed on instances
+type PlacementSnapshot interface {
+	// Instances returns the placement instances
+	Instances() []PlacementInstance
+
+	// SetInstances sets the placement instances
+	SetInstances(insts []PlacementInstance) PlacementSnapshot
+
+	// Replication returns the placement replication description or nil if none
+	Replication() ServiceReplication
+
+	// SetReplication sets the placement replication description or nil if none
+	SetReplication(r ServiceReplication) PlacementSnapshot
+
+	// Sharding returns the placement sharding description or nil if none
+	Sharding() ServiceSharding
+
+	// SetSharding sets the placement sharding description or nil if none
+	SetSharding(s ServiceSharding) PlacementSnapshot
+
+	// Validate checks if the placement is valid
+	Validate() error
+}
 
 // ServiceReplication describes the replication of a service
 type ServiceReplication interface {
@@ -65,9 +82,6 @@ type ServiceReplication interface {
 	SetReplicas(r int) ServiceReplication
 }
 
-// NewServiceReplication creates a new ServiceReplication
-func NewServiceReplication() ServiceReplication { return new(serviceReplication) }
-
 // ServiceSharding describes the sharding of a service
 type ServiceSharding interface {
 	// NumShards is the number of shards to use for sharding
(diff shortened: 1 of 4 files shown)


services/services.go
```diff
@@ -53,8 +49,29 @@ type Service interface {
 	SetSharding(s ServiceSharding) Service
 }
 
-// NewService creates a new Service
-func NewService() Service { return new(service) }
+// PlacementSnapshot describes how shards are placed on instances
+type PlacementSnapshot interface {
```
robskillington: `ServicePlacement`?

services/services.go
```diff
@@ -53,8 +49,29 @@ type Service interface {
...
+	SetReplication(r ServiceReplication) PlacementSnapshot
+
+	// Sharding returns the placement sharding description or nil if none
+	Sharding() ServiceSharding
+
+	// SetSharding sets the placement sharding description or nil if none
+	SetSharding(s ServiceSharding) PlacementSnapshot
+
+	// Validate checks if the placement is valid
+	Validate() error
```
robskillington: Do we need `Validate()` anymore if this is just a plain old object or POGO one might say?

services/services.go
```diff
@@ -77,25 +91,27 @@ type ServiceSharding interface {
...
-	Endpoint() string                         // Endpoint address for contacting the instance
-	SetEndpoint(e string) ServiceInstance     // sets the endpoint address for the instance
-	Shards() shard.Shards                     // Shards owned by the instance
-	SetShards(s shard.Shards) ServiceInstance // sets the Shards assigned to the instance
+	PlacementInstance() string
+	SetPlacementInstance(i PlacementInstance) ServiceInstance
+	Service() string                         // the service implemented by the instance
+	SetService(s string) ServiceInstance     // sets the service implemented by the instance
+	Endpoints() []string                     // Endpoint addresses for contacting the instance
+	SetEndpoints(e []string) ServiceInstance // sets the endpoint addresses for the instance
```
robskillington: Let's keep a single endpoint per service instance as discussed.

services/services.go
```diff
@@ -109,111 +125,52 @@ type Advertisement interface {
...
-	SetZones(zones []string) QueryOptions    // sets the list of zones to consult
 	IncludeUnhealthy() bool                  // if true, will return unhealthy instances
 	SetIncludeUnhealthy(h bool) QueryOptions // sets whether to include unhealthy instances
 }
 
-// NewQueryOptions creates new QueryOptions
-func NewQueryOptions() QueryOptions { return new(queryOptions) }
-
 // Services provides access to the service topology
 type Services interface {
```
robskillington: Let's add a method as discussed that returns a `PlacementService` for a given `query ServiceQuery` then each parameter on `PlacementService` already has the context of a service.

services/services.go
```diff
@@ -109,111 +125,52 @@ type Advertisement interface {
...
+// all write or update operations will persist the generated snapshot before returning success
+type PlacementService interface {
+	BuildInitialPlacement(service ServiceQuery, hosts []PlacementInstance, shardLen int, rf int) (PlacementSnapshot, error)
+	AddReplica(service ServiceQuery) (PlacementSnapshot, error)
+	AddHost(service ServiceQuery, candidateHosts []PlacementInstance) (PlacementSnapshot, error)
+	RemoveHost(service ServiceQuery, host PlacementInstance) (PlacementSnapshot, error)
+	ReplaceHost(service ServiceQuery, leavingHost PlacementInstance, candidateHosts []PlacementInstance) (PlacementSnapshot, error)
+
+	// Snapshot gets the persisted snapshot for service
+	Snapshot(service ServiceQuery) (PlacementSnapshot, error)
```
robskillington: As discussed, would be good to have this `PlacementService` operate under the context of a given service already.

shard/shard.go
```diff
@@ -20,10 +20,24 @@
 
 package shard
 
+const (
+	// INITIALIZING represents a shard newly assigned to an instance
+	INITIALIZING State = iota
```
robskillington: Let's stick with our current const naming scheme, this should just be `Initializing`.

shard/shard.go
```diff
@@ -20,10 +20,24 @@
...
+
+	// AVAILABLE represends a shard bootstraped and ready to serve
+	AVAILABLE
```
robskillington: Let's stick with our current const naming scheme, this should just be `Available`.

services/services.go
cw9: I'd still have Validate() here, cause we want to make sure every placement is validate before we save it, it's either here or a "validate(p ServicePlacement) error" somewhere in the persistent layer.

shard/shard.go
cw9: will fix

services/services.go
cw9: sure

PR #39, opened 2016-11-23
Support shard state in placement operations

cw9:
@xichen2020 @robskillington


services/placement/algo/helper.go
```diff
@@ -296,32 +317,39 @@ func (ph placementHelper) getShardLen() int {
...
 	}
 
-	from.Shards().RemoveShard(shard)
-	delete(ph.shardToInstanceMap[shard], from)
+	if s.State() == shard.Initializing {
+		from.Shards().Remove(s.ID())
+	} else if s.State() == shard.Available {
+		s.SetState(shard.Leaving)
+	} else if s.State() == shard.Leaving {
+		panic("should never move a leaving shard")
```
robskillington: Some of this code will likely one day run inside a discovery service no?  If so I think the correct thing here is to return an error and abandon a change rather than panic and bring down a running service.

robskillington: Also:  how does this shard ever get removed btw once the shard has finished leaving?

services/placement/algo/helper.go
```diff
@@ -338,7 +366,12 @@ func newHeap(
 	targetLoad map[string]int,
 	rackToWeightMap map[string]uint32,
 ) *instanceHeap {
-	hHeap := &instanceHeap{capacityAscending: capacityAscending, instances: instances, targetLoad: targetLoad, rackToWeightMap: rackToWeightMap}
+	hHeap := &instanceHeap{
```
robskillington: Perhaps name the local variable here just `h`?  To avoid the strange `hHeap` name?

services/placement/algo/helper.go
```diff
@@ -385,16 +418,79 @@ func (h *instanceHeap) Pop() interface{} {
 	return instance
 }
 
-func addInstanceToPlacement(p services.ServicePlacement, i services.PlacementInstance) (services.ServicePlacement, services.PlacementInstance, error) {
+// MarkShardAvailable marks the state of a shard to available
+func MarkShardAvailable(p services.ServicePlacement, instanceID string, shardID uint32) (services.ServicePlacement, error) {
```
robskillington: Should this method perhaps live on the `PlacementService`?  It's not extremely discoverable as a method sitting in the `algo` package no?

services/placement/algo/helper.go
```diff
@@ -416,13 +512,71 @@ func copyPlacement(p services.ServicePlacement) services.ServicePlacement {
...
+	return placement.NewInstance().
+		SetID(instance.ID()).
+		SetRack(instance.Rack()).
+		SetZone(instance.Zone()).
+		SetWeight(instance.Weight()).
+		SetEndpoint(instance.Endpoint()).
+		SetShards(copyShards(instance.Shards()))
+}
+
+func copyShards(shards shard.Shards) shard.Shards {
```
robskillington: Perhaps `cloneShards` is a better name?  Copy named methods usually take a destination and a source.

services/placement/placement.go
```diff
@@ -83,27 +82,43 @@ func (p placement) NumShards() int {
...
+			if !exist {
+				return errUnexpectedShards
+			}
+			if s.State() == shard.Available {
+				shardCountMap[s.ID()] = count + 1
+				totalCapacity++
+			} else if s.State() == shard.Initializing {
+				totalInit++
+				shardCountMap[s.ID()] = count + 1
+				totalCapacity++
```
robskillington: If there is a copy of this shard already available, should it not count towards the `totalCapacity`?

Otherwise `expectedTotal != totalCapacity`.

e.g. imagine a case of bringing a new node online:

```
# stage 1
hostA { shardRange = 0..1023 }
hostB { shardRange = nil }

# stage 2
hostA { shardRange = 0..1023 }
hostB { shardRange = 512..1023 }

# stage 3
hostA { shardRange = 0..511 }
hostB { shardRange = 512..1023 }
```

For this case, in stage 2, the totalCapacity would be 1536 and would not equal the expectedTotal of `1024 * 1 = 1024`.

services/placement/service/placementservice.go
```diff
@@ -84,29 +77,47 @@ func (ps placementService) BuildInitialPlacement(
...
 	if p, err = ps.algo.AddReplica(p); err != nil {
 		return nil, err
 	}
 
-	return p, ps.validateAndSaveSnapshot(p)
+	if err := placement.Validate(p); err != nil {
+		return nil, err
+	}
+
+	// TODO(chaowant) this will be removed once the m3db nodes start to mark shards as available
```
robskillington: For this case, let's just assume for now that nodes will make them as available.  That is how we intend to launch with this feature anyhow.  i.e. perhaps remove the `markAllShardsAsAvailable(...)` call for this case.

services/placement/service/placementservice.go
```diff
@@ -118,40 +129,56 @@ func (ps placementService) AddInstance(
...
+	p, err = markAllShardsAsAvailable(p)
```
robskillington: Yeah we should not mark these shards as available, we need to rely on nodes marking them when they are initialized.

services/placement/service/placementservice.go
```diff
@@ -161,15 +188,68 @@ func (ps placementService) ReplaceInstance(
...
```
robskillington: This too, on replace the nodes will mark the shards as available once initialized.

shard/shard.go
```diff
@@ -20,40 +20,78 @@
...
 package shard
 
+const (
+	// Initializing represents a shard newly assigned to an instance
+	Initializing State = iota
+
+	// Available represends a shard bootstraped and ready to serve
+	Available
+
+	// Leaving is laeving
```
robskillington: Typo here, also perhaps make it more similar to the other comments: "Leaving represents a shard that is intending to be removed"

shard/shard.go
```diff
@@ -20,40 +20,78 @@
 
 package shard
 
+const (
+	// Initializing represents a shard newly assigned to an instance
+	Initializing State = iota
+
+	// Available represends a shard bootstraped and ready to serve
```
robskillington: Typo here, "represends"

shard/shard.go
```diff
@@ -20,40 +20,78 @@
...
+	Available
+
+	// Leaving is laeving
+	Leaving
+)
+
+// State represends the state of a shard
```
robskillington: Typo here, "represends"

shard/shard.go
```diff
@@ -90,23 +119,40 @@ func (s shards) NumShards() int {
...
 	delete(s.shardsMap, shard)
 }
 
-func (s shards) ContainsShard(shard uint32) bool {
+func (s shards) Contains(shard uint32) bool {
 	_, ok := s.shardsMap[shard]
 	return ok
 }
+
+func (s shards) NumShardsInState(state State) int {
```
robskillington: Perhaps slightly more idiomatic naming here would be `NumShardsForState(...)` and `ShardsForState(...)`?

services/placement/algo/helper.go
cw9: The leaving shard will be removed in MarkAsAvailable, when the Initializing shard getting marked as Available, this Leaving shard will be removed.

the panic condition should never be called in this algorithm, I could return an error for that case, but all the callers will have to handle that error which is not expected to happen, I could also just remove that else if branch.

services/placement/algo/helper.go
cw9: sure will do

services/placement/algo/helper.go
cw9: sure, will do

services/placement/service/placementservice.go
```diff
@@ -84,29 +77,47 @@ func (ps placementService) BuildInitialPlacement(
 		}
 	}
 
-	return p, ps.validateAndSaveSnapshot(p)
+	if err := placement.Validate(p); err != nil {
+		return nil, err
+	}
+
+	// TODO(chaowant) this will be removed once the m3db nodes start to mark shards as available
```
cw9: I could make init placement to have them as available, I'm fine either way, I just felt that "Initializing" is probably more accurate in describing the state of that shard

services/placement/service/placementservice.go
cw9: Actually the reason I added this is so that I can land this diff before your resharding diff. I felt it is safer to decouple the 2 features. Then you need to remove these code when you start to call MarkShardAvaialbe in M3DB

shard/shard.go
cw9: you mean "represents"? will fix

shard/shard.go
cw9: yeah lol I had this comment for lint while coding, forgot to edit it afterwards

shard/shard.go
cw9: sure, will do

services/placement/algo/helper.go
robskillington: Hey - so yeah because this algorithm will execute as service code we need to bubble up errors, no matter how unlikely.  Someone may come in and introduce a bug that triggers this case.  If so it is better for the functionality to not work when they deploy to the service rather than crashing the entire service.

We should bubble this error up.

services/placement/algo/helper.go
robskillington: I didn't realize that the `PlacementService` calls into this method from it's own methods.  Seems fair enough.

services/placement/placement.go
robskillington: Ah I see, makes sense.  Thanks for the explanation.

services/placement/algo/helper.go
cw9: sure, will do

services/placement/algo/helper.go
```diff
@@ -67,21 +74,26 @@ func (ph *placementHelper) MoveOneShard(from, to services.PlacementInstance) boo
...
+		return false
 	}
-	return false
+
+	newShard := shard.NewShard(s.ID()).SetState(shard.Initializing)
+	if from != nil && s.State() != shard.Initializing {
+		newShard = newShard.SetSourceID(from.ID())
+	}
+	ph.assignShardToInstance(newShard, to)
+	ph.removeShardFromInstance(s, from)
```
robskillington: Should check if this returns an error and if so return false here no?

cw9: ah, yeah, will fix

PR #41, opened 2016-12-04
Rely on external callers marking shards available

robskillington:
cc @cw9 @xichen2020 @martin-mao @ben-lerner @prateek 

Outstanding question for @cw9 : when do the leaving shards get removed and then also when is instance completely taken out of the placement?  What calls to the `PlacementService` achieve this?

### DIFF
--- a/services/placement/service/placementservice_test.go
+++ b/services/placement/service/placementservice_test.go
@@ -42,8 +42,8 @@ func TestGoodWorkflow(t *testing.T) {
 	p := NewPlacementService(NewMockStorage(), testServiceID(), placement.NewOptions())
 	testGoodWorkflow(t, p)
 
-	// p = NewPlacementService(NewMockStorage(), testServiceID(), placement.NewOptions().SetLooseRackCheck(true))
-	// testGoodWorkflow(t, p)
+	p = NewPlacementService(NewMockStorage(), testServiceID(), placement.NewOptions().SetLooseRackCheck(true))
+	testGoodWorkflow(t, p)
 }
 
 func testGoodWorkflow(t *testing.T, p services.PlacementService) {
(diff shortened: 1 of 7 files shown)


cw9:
You can call either MarkShardAvailable or MarkInstanceAvailable, MarkShardAvailable needs to be called per shard, MarkInstanceAvailable will be marking all Initializing shards on the instance as Available. Leaving shards will be removed from instance once its corresponding shard is being marked as Available. Once an instance has no shards in any state left on it, it will also be removed from the placement

services/placement/service/placementservice_test.go
```diff
@@ -702,3 +718,27 @@ func (ms *mockStorage) Placement(service services.ServiceID) (services.ServicePl
...
+	for _, i := range p.Instances() {
+		if len(i.Shards().ShardsForState(shard.Initializing)) == 0 {
+			continue
+		}
+		err := ps.MarkInstanceAvailable(i.ID())
+		require.NoError(t, err)
+	}
+}
+
+func placementToString(
```
cw9: still need this?

robskillington: It's useful to have around, I'd probably end up re-writing it next time I come back to these tests.  Perhaps it'd be better to put it as a `String()` method on `ServicePlacement` however.

I might do that unless you have objections?

services/placement/service/placementservice_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/m3db/m3cluster/services/placement"
 	"github.com/m3db/m3cluster/shard"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
robskillington: `require` is good for if you want the test to bailout without running the rest of the checks.  There are some scenarios where that's more desirable (i.e. would cause a nil panic later in the test, hence better to use a require above it to ensure that doesn't happen). 

Assert is more preferable if it won't crash however as you get more signal back about what things are failing and where when tests start to fail.

services/placement/service/placementservice_test.go
cw9: sure, sounds good to me

services/placement/service/placementservice_test.go
cw9: kk

cw9:
LGTM

PR #42, opened 2016-12-07
Carry sourceID when moving an Initializing shard

cw9:
Fix an edge case caused by:
1, add a node, so it will be assigned with some shards in Initializing state, with sourceID to reference where these shards come from.
2, remove this node before it finish bootstrapping it's shards
3, the Initializing shard will be moved to another node but the sourceID got lost.

@robskillington

### DIFF
--- a/services/placement/algo/helper.go
+++ b/services/placement/algo/helper.go
@@ -71,9 +71,6 @@ func (ph *placementHelper) TargetLoadForInstance(id string) int {
 }
 
 func (ph *placementHelper) MoveOneShard(from, to services.PlacementInstance) bool {
-	if from == nil {
-		return false
-	}
 	for _, s := range from.Shards().All() {
 		if s.State() != shard.Leaving && ph.MoveShard(s, from, to) {
 			return true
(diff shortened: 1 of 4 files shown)


services/placement/algo/helper.go
```diff
@@ -71,9 +71,6 @@ func (ph *placementHelper) TargetLoadForInstance(id string) int {
 }
 
 func (ph *placementHelper) MoveOneShard(from, to services.PlacementInstance) bool {
-	if from == nil {
-		return false
-	}
```
robskillington: Is there a reason to remove this?  Or was it overly defensive programming?

services/placement/algo/helper.go
```diff
@@ -293,6 +291,9 @@ func (ph *placementHelper) scanCurrentLoad() {
 		totalWeight += h.Weight()
 
 		for _, s := range h.Shards().All() {
+			if s.State() == shard.Leaving {
+				continue
+			}
```
robskillington: Ah this is to avoid leaving shards looking like "load" on an instance?

services/placement/algo/helper.go
cw9: it was overly defensive, no caller would call it on a nil from

services/placement/algo/helper.go
cw9: yes, there are test cases making sure it's gone

robskillington: Ah right, I see.

services/placement/algo/helper.go
robskillington: Sounds good yeah.

PR #43, opened 2016-12-13
Fill in Endpoint field when adding instance

cw9:
@robskillington

### DIFF
--- a/services/placement/algo/algo_test.go
+++ b/services/placement/algo/algo_test.go
@@ -32,17 +32,17 @@ import (
 )
 
 func TestGoodCase(t *testing.T) {
-	h1 := placement.NewEmptyInstance("r1h1", "r1", "z1", 1)
-	h2 := placement.NewEmptyInstance("r1h2", "r1", "z1", 1)
-	h3 := placement.NewEmptyInstance("r2h3", "r2", "z1", 1)
-	h4 := placement.NewEmptyInstance("r2h4", "r2", "z1", 1)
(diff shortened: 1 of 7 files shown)


services/placement/algo/algo_test.go
```diff
@@ -705,28 +709,30 @@ func TestAddInstance(t *testing.T) {
...
-	i2 := placement.NewEmptyInstance("i2", "r2", "", "endpoint", 1)
+	i2 := placement.NewEmptyInstance("i2", "r2", "", "e2", 1)
 	p, err := a.AddInstance(p, i2)
 	assert.NoError(t, err)
 	assert.Equal(t, 1, p.ReplicaFactor())
 	i1 = p.Instance("i1")
 	i2 = p.Instance("i2")
 	assert.Equal(t, 1, loadOnInstance(i1))
 	assert.Equal(t, 2, i1.Shards().NumShards())
+	assert.Equal(t, "e1", i1.Endpoint())
```
robskillington: Shouldn't we be checking the endpoint coming back from `AddInstance(..)`?  Otherwise we don't test that it got cloned properly.

e.g. `p.Instance(i1.InstanceID()).Endpoint()`?

cw9: yeah i1 is from the placement after AddInstance() call, see line 716

robskillington: Indeed it is, I am blind!  Thankyou.

PR #44, opened 2016-12-14
Modify Instance() interface

cw9:
@xichen2020 @robskillington


services/placement/placement.go
```diff
@@ -59,13 +59,13 @@ func (p placement) NumInstances() int {
 	return len(p.instances)
 }
 
-func (p placement) Instance(id string) services.PlacementInstance {
+func (p placement) Instance(id string) (services.PlacementInstance, bool) {
 	for _, instance := range p.Instances() {
```
robskillington: @cw9 wouldn't it be most efficient to construct a map of all the instances by ID and then return just:
```go
instance, ok := p.instancesByID[id]
return instance, ok
```

We do a similar thing in the static topology map in M3DB to speed up lookups.

services/placement/algo/helper.go
```diff
@@ -490,8 +490,8 @@ func addInstanceToPlacement(p services.ServicePlacement, i services.PlacementIns
 }
 
 func removeInstanceFromPlacement(p services.ServicePlacement, id string) (services.ServicePlacement, services.PlacementInstance, error) {
-	leavingInstance := p.Instance(id)
-	if leavingInstance == nil {
+	leavingInstance, exist := p.Instance(id)
+	if !exist {
 		return nil, nil, errInstanceAbsent
```
xichen2020: hm might be better to return the same error in `MarkShardAvailable` if the instance is not found?

PR #45, opened 2016-12-14
Modify ConfigServiceClient interface

cw9:
Some config service client user only needs a KV client or a Services
client, not both. This interface change allows us to create those
clients lazily.


cw9:
@xichen2020 @robskillington

PR #46, opened 2016-12-18
KV client based on etcd

cw9:
@xichen2020 @robskillington


kv/etcd/etcd.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/m3db/m3x/log"
+	"github.com/m3db/m3x/retry"
+	"golang.org/x/net/context"
+)
+
+var noopCancel func()
+
+// NewStore creates a kv store based on etcd
+func NewStore(opts Options) (kv.Store, error) {
+	etcd, err := clientv3.New(clientv3.Config{
```
robskillington: Might it be worth having callers pass in the `etcd` client so they can configure it however they may like?  e.g. `NewStore(client *clientv3.Client, opts Options) (kv.Store, error)`

Just in case in there is something they would prefer to initialize themselves and/or pass in a shared client they already have?

Also: we could go one step further and define the interface that matches methods of `*clientv3.Client` so that technically you could pass a mock etcd client in here, especially for testing purposes this would be most ideal?

We could define the following interface that matches the `*clientv3.Client` methods so it can be seamlessly passed in:

```go
type EtcdClient interface {
  clientv3.KV
  clientv3.Watcher

  Close() error
  Ctx() context.Context
  Endpoints() (eps []string)
  SetEndpoints(eps ...string)
  Sync(ctx context.Context) error
  Dial(endpoint string) (*grpc.ClientConn, error)
  ActiveConnection() *grpc.ClientConn
}
```

kv/etcd/etcd.go
```diff
@@ -0,0 +1,220 @@
...
+	r, err := c.kv.Get(ctx, c.opts.KeyFn()(key))
+	if err != nil {
+		return nil, err
+	}
+
+	if r.Count == 0 {
+		return nil, kv.ErrNotFound
+	}
+
+	if r.Count > 1 {
```
robskillington: Curious, in what case would `Count > 1`?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,220 @@
...
+			clientv3.WithProgressNotify(),
+			// Receive initial notification once the watch channel is created
+			clientv3.WithCreatedNotify(),
+		)
+
+		watchable = kv.NewValueWatchable()
+		c.watchables[key] = watchable
+
+		go func() {
+			for r := range watchChan {
```
robskillington: Will closing the returned watchable cause this `for` loop to break?  Or will this goroutine never be able to be closed and "leak" in a sense?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,220 @@
...
+func (c *client) Set(key string, v proto.Message) (int, error) {
+	ctx, cancel := c.context()
+	defer cancel()
+
+	value, err := proto.Marshal(v)
+	if err != nil {
+		return 0, err
+	}
+
+	r, err := c.kv.Put(ctx, c.opts.KeyFn()(key), string(value), clientv3.WithPrevKV())
```
robskillington: Hm might not be great to store the raw bytes as `string` here, `etcd` has an HTTP API and it may not play nicely with this API to use raw bytes as string.  Probably more prudent to base64 the bytes into a string, then decoded base64 before unmarshalling when retrieving a value.

kv/store.go
```diff
@@ -21,77 +21,31 @@
...
-	"github.com/m3db/m3x/watch"
+	xwatch "github.com/m3db/m3x/watch"
 )
 
-var (
-	// ErrVersionMismatch is returned when attempting a CheckAndSet and the
-	// key is not at the provided version
-	ErrVersionMismatch = errors.New("key is not at the specified version")
+type value struct {
+	CurVersion int    `json:"version"`
```
robskillington: Any disposition to just naming this `Version` as opposed to `CurVersion`?

kv/etcd/etcd.go
cw9: Sounds good, will refactor the interface a bit

kv/etcd/etcd.go
cw9: The Get() in etcd can be configured to get all keys with a prefix or all keys in a certain range, although we are not passing in those options in the Get() call here, just being defensive

kv/etcd/etcd.go
cw9: We are returning the watch, not the watchable, multiple watches could be sharing the same watchable. But if all the watches closed the watchable will be still kept alive and being updated in the background, we could close the watchable once no watches are subscribing to the watchable, but I don't feel super necessary here, keeping it alive could make next subscriber easier.

kv/store.go
cw9: Unfortunately Version conflicts with the function name Version()

kv/etcd/etcd.go
cw9: hmm, can do that, but the doc: https://github.com/coreos/etcd/blob/master/clientv3/kv.go#L36 says can just do string(), probably that's safe?

kv/etcd/etcd.go
cw9: Also we don't need to mock the interface, they provided an integration package that can start up an simple etcd test cluster, which I used in unit test

kv/etcd/etcd.go
robskillington: Ah I see, sounds good 👍

kv/etcd/etcd.go
robskillington: Hm, I'm worried about leaking however.  Doesn't seem clean to have a goroutine forever around after that you can never close.

kv/etcd/etcd.go
robskillington: Oh interesting, well in that case we can probably keep it.  Seems like a bad API to make you perform `string(...)` on a bytes slice.. but if that's how it is then that's how it is.

kv/store.go
robskillington: Ah I see.

robskillington:
Only thing still concerning is just the potential goroutine leak.

kv/etcd/etcd.go
cw9: Will put up a diff for this

kv/etcd/etcd.go
xiang90: string is immutable []byte in go

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+				case r, ok := <-watchChan:
+					if ok {
+						c.processNotification(r, watchable, key)
+					} else {
+						c.logger.Warnf("etcd watch channel closed on key %s, recreating a watch channel", key)
+
+						// avoid recreating watch channel too frequently
+						time.Sleep(c.opts.WatchChanResetInterval())
+
+						watchChan = c.watcher.Watch(
```
cw9: I don't necessarily need to get every event, I just want to get the latest version and get it soon, I think the watch + WithProgressNotify() should satisfy my needs here. 
Yeah I agree I barely could hit here, just being defensive, because we might deploy this client to lots of hosts, even if just a few of them end up with a closed channel we can't afford to restart them all.

xiang90: OK. If you only want the latest result, then this is fine. 

Can you explain why do you need to set WithProgressNotify? It is usually only needed when you want to do fast resume or detect not responding server. ProgressNotify will notify the watcher its current progress (global revision) even if nothing interesting happened to the watcher. For example, you watch on key `foo` for 10 minutes without progressNotify, if nothing happened then etcd will send nothing to the watcher. If you watch with progressNotify in this case, etcd will send you an empty response with current revision. See this (https://github.com/coreos/etcd/blob/master/etcdserver/etcdserverpb/rpc.proto#L601-L605) for more details.

cw9: Yeah that's exactly what I wanted, I'm trying be defensive for the case that Get() call followed by the watch notification failed with all retries on a watch update, without WithProgressNotify, I'll wait until next event happens on the key for this go routine to try another Get(), which could be really long. now with WithProgressNotify, I can at least try another Get() every 10 mins, which does not add much work load but make sure it has another chance to check for any updates every 10 mins.

As you can see I use watch updates as notifications only, rather than looking into what kind of update it is, be it update or create or delete. Is there any option on the Watch call to not get those information, only get notifications? If not, probably worth adding one?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/m3db/m3x/log"
+	"github.com/m3db/m3x/retry"
+	"github.com/uber-go/tally"
+	"golang.org/x/net/context"
+)
+
+var noopCancel func()
+
+// NewStore creates a kv store based on etcd
+func NewStore(etcd *clientv3.Client, opts Options) kv.Store {
```
xichen2020: `client` might be a better name than `etcd` here?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (c *client) Watch(key string) (kv.ValueWatch, error) {
+	c.Lock()
+	watchable, ok := c.watchables[key]
+	if !ok {
+		watchChan := c.watcher.Watch(
+			context.Background(),
+			c.opts.KeyFn()(key),
+			clientv3.WithProgressNotify(),
```
xichen2020: hm why do we need this?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+		c.watchables[key] = watchable
+
+		go func() {
+			ticker := time.Tick(c.opts.WatchChanCheckInterval())
+
+			for {
+				select {
+				case r, ok := <-watchChan:
+					if ok {
+						c.processNotification(r, watchable, key)
```
xichen2020: Presumably we won't miss an event if an update is received while we are processing the notification?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	r, err := c.kv.Put(ctx, c.opts.KeyFn()(key), string(value), clientv3.WithPrevKV())
+	if err != nil {
+		c.m.etcdPutError.Inc(1)
+		return 0, err
+	}
+
+	// if there is no prev kv, means this is the first version of the key
+	if r.PrevKv == nil {
```
xichen2020: What happens if the previous kv gets compacted?

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+	defer cancel()
+
+	value, err := proto.Marshal(v)
+	if err != nil {
+		return 0, err
+	}
+
+	key = c.opts.KeyFn()(key)
+	r, err := c.kv.Txn(ctx).
+		If(clientv3.Compare(clientv3.Version(key), "=", 0)).
```
xichen2020: Might be better to make `0` a const

kv/etcd/etcd.go
```diff
@@ -0,0 +1,292 @@
...
+		c.m.etcdTnxError.Inc(1)
+		return 0, err
+	}
+	if !r.Succeeded {
+		return 0, kv.ErrAlreadyExists
+	}
+	return 1, nil
+}
+
+func (c *client) CheckAndSet(key string, version int, v proto.Message) (int, error) {
```
xichen2020: Looks like this is a generic version of `SetIfNotExists`? If so maybe we should change `SetIfNotExists` to
simply call `CheckAndSet(key, 0, v)` instead?

kv/etcd/options.go
```diff
@@ -0,0 +1,167 @@
...
+		SetInstrumentsOptions(instrument.NewOptions()).
+		SetRetryOptions(xretry.NewOptions()).
+		SetWatchChanCheckInterval(defaultWatchChanCheckInterval).
+		SetWatchChanResetInterval(defaultWatchChanResetInterval).
+		SetKeyFn(defaultKeyFn)
+}
+
+func (o options) Validate() error {
+	if o.iopts == nil {
+		return errors.New("no instrument options")
```
xichen2020: nit: might be better to make these `vars` instead

kv/etcd/etcd.go
cw9: unfortunately that would conflict with the client struct

kv/etcd/etcd.go
cw9: I explained in the conversation below, basically that lets us do another check on the latest value in kv store every 10 mins.

kv/etcd/etcd.go
cw9: hmm, I don't think it would.

kv/etcd/etcd.go
cw9: yeah sounds good.

kv/etcd/etcd.go
cw9: sure can do

kv/etcd/options.go
cw9: yeah they are only used once here, I really don't want to assign them each with a name but only use it once unless we are really benefitting from it.

kv/etcd/etcd.go
cw9: yeah good catch, in that case I have to make another Get() call to get the version

cw9: actually, the previous kv won't be compacted since this set will only inc the revision by 1, unless we compact every version we won't miss it.

kv/etcd/etcd.go
cw9: added a test for it

kv/etcd/etcd.go
xiang90: > Is there any option on the Watch call to not get those information, only get notifications? If not, probably worth adding one?

No. This is not the exact use case we designed the watch API for (the current API is richer in feature). If you worry about the additional cost by sending the events, can you create an issue in etcd repo, explaining your use case and letting the etcd team know why you need the option?

Thanks.

kv/etcd/etcd.go
xichen2020: hm it shouldn't, it's a locally scoped variable name rather than a type name, so it *should* work

kv/etcd/etcd.go
xichen2020: This seems somewhat hackish, basically we are using WithProgressNotify to bound the maximum staleness of the keys on the client side, which is not quite reliable (e.g., server may determine when to send depending on the server load etc). If we are worried that `Get` could fail even with retries, can we simply increase the number of retries with an upper bound on the maximum backoff delay (e.g., 10 minutes)?

kv/etcd/etcd.go
xichen2020: Yeah fair enough, although i could see a edge case where this could happen (e.g., two clients setting the same version at the same time, the faster client bumps the version, then compaction kicks in, then the slower client won't see the previous versions). Is there a way to distinguish compacted versions and no versions?

kv/etcd/options.go
xichen2020: Sounds good

kv/etcd/etcd.go
cw9: because when you do client.KV, the compiler does not know if the client is the clientv3.Client or the local struct

kv/etcd/etcd.go
cw9: I'm not trying to bound the staleness, rather I'm trying to use it as a cheap passive periodical check in case the watch became sad, like disconnected from the server. I would also add a bit more retries to the default options but I don't want to retry too much because it's going to block the thread and delaying the check of number of watches on this key, and delaying deleting the watch channel if no one watches on it anymore.

kv/etcd/etcd.go
cw9: You can't set the same version with the Set API, if 2 request came in, the fast one will inc the version v0 to v0+1 first then the second one will inc it from v0+1 to v0+2, they will be able to see different previous versions(v0 and v0+1) and they will surely see them because they are both only 1 revision old. 

The CAS API is for setting the value based on a certain version, where the atomicity will be guaranteed by using transaction

kv/etcd/etcd.go
xichen2020: Ah I see, you are returning a `client` object here, yeah that won't work. Maybe just name it `c` then?

kv/etcd/etcd.go
xichen2020: Okay in that case can we add some comments to explain why it's necessary?

kv/etcd/etcd.go
xichen2020: Ah yeah the API doesn't allow you to specify the version, fair enough

kv/etcd/etcd.go
cw9: sure, will do

PR #47, opened 2016-12-20
No longer pin m3x to a certain version in glide

cw9:
Pinning m3x to a version seems unnecessary to me

@xichen2020 @robskillington

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: ae5c127a48fafa2da20ddc41151e5a108e7f3254074adf3ab66ff285b7636afd
-updated: 2016-09-01T14:12:12.96861304-04:00
+hash: 29e2df8267349dba733160bef18d3e0fe56958f96d8170e49b576dcdf00875fa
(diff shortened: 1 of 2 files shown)


xichen2020:
hm why? If we don't pin it, every time we run glide update it'll automatically pick up the latest, which may not be what we want

cw9:
Hmm, we should make sure master of m3x always work no? What I want now is to add instrument subpackage to it, I could just add that, but felt we don't need to pin m3x version thus removed the whole pin.

xichen2020:
Yeah, I'm more worried about the scenario where m3cluster relies on an old version of m3x, which works just fine, and the tip of master in m3x has some backward-incompatible changes, so if we update the deps of m3cluster, we'll pick those changes up, which we may or may not want to (especially in the situation of emergency deploys)

cw9:
Yeah, that's true, I'll just add a subpackage and use current tip of master with it

cw9:
Actually I'll just do the glide update in the pr with code change

PR #48, opened 2016-12-20
Services client based on kv store

cw9:
Moved the sd client code out from statsdex
@xichen2020 @robskillington


cw9:
Just to have a SHA to test in statsdex, will post a real pr later

xichen2020:
Can just push a branch without opening a PR =p

cw9:
yeah, forgot that :(

PR #50, opened 2017-01-01
Add heartbeat store and use it in Services client

cw9:
added APIs to get/set ServiceMetadata
implemented advertise/unadvertise based on heartbeat store and ServiceMetadata
support requests with IncludeUnhealthy = false in Query() and Watch() 
added test for services client

@xichen2020 @robskillington


services/heartbeat/etcd/etcd.go
```diff
@@ -0,0 +1,78 @@
...
+	kv clientv3.KV
+}
+
+func (c *client) Heartbeat(service, id string, ttl time.Duration) error {
+	resp, err := c.l.Grant(context.Background(), int64(ttl/time.Second))
+	if err != nil {
+		return err
+	}
+
+	_, err = c.kv.Put(context.Background(), heartbeatKey(service, id), "", clientv3.WithLease(resp.ID))
```
xiang90: why not keepalive the lease if the key is already attached to a lease? A lease grant + put is expensive.

cw9: how expensive is the grant + put?

I didn't do that because I want to allow different ttl for each heartbeat and I can't have multiple keys share the same lease, so I didn't think there is too much value keeping track of the lease id and do keepalive on it unless it's really expensive to do the grant+put way.

xiang90: two rounds of consensus protocol vs one memory operation. how often do you change TTL duration? if it is not often (and it really should not normally), keepalive should be used instead.

cw9: The ttl will not change often, but one main use case I plan to support is to have a few service nodes taking requests from callers and then call this API to store the heartbeats in etcd, if the number of nodes is larger than ttl/heartbeating frequency then there is a big chance the cached old lease on a node will be outdated when the caller reaches the node again, so I'd still have to do a grant+put. 

We don't plan to have a lot of heartbeating for now and I assume the KeepAliveOnce will go through one round of consensus too? in that case we will only be saving 1 round of consensus * number of heartbeats, which is not a lot. Your concern is definitely appreciated, and if the performance on this piece of code became an issue I will consider replacing it with KeepAlive.

cw9: also is it safe to just use ttl/time.Second as the number of seconds here? or does it need a lower bound?

cw9: Got it, thank you for the clarification!

services/client/services.go
```diff
@@ -96,129 +193,257 @@ func (s *client) Query(sid services.ServiceID, opts services.QueryOptions) (serv
...
-			s.opts.InstrumentsOptions().Logger().Errorf("could not unmarshal update from kv store for cluster")
-			errCounter.Inc(1)
-			continue
+	var err error
+
+	for {
+		select {
+		case <-vw.C():
+			value := vw.Get()
+			c.opts.InstrumentsOptions().Logger().Infof("received topology update notification on version %d", value.Version())
```
xichen2020: nit: can do `log := c.opts.InstrumentsOptions().Logger()` outside the for loop so you can directly do `log.Infof(...)` and below

services/client/services.go
```diff
@@ -96,129 +193,257 @@ func (s *client) Query(sid services.ServiceID, opts services.QueryOptions) (serv
...
+			if !equalIDs(ids, newIDs) {
+				ids = newIDs
+				w.Update(filterHealthyInstances(service, ids, qopts.IncludeUnhealthy()))
+			}
 		}
-		s.opts.InstrumentsOptions().Logger().Infof("successfully parsed placement on version %d", value.Version())
+	}
+}
+
+func isHealthy(ad services.Advertisement) bool {
```
xichen2020: Could be simplified to
```
healthFn := ad.Health()
return healthFn == nil || healthFn() == nil
```

services/client/services.go
```diff
@@ -269,26 +496,48 @@ func newServiceDiscoveryMetrics(m tally.Scope) serviceDiscoveryMetrics {
...
 }
 
-func validateRequest(sid services.ServiceID) error {
-	if sid.Environment() == "" {
-		return errNoEnv
-	}
-	if sid.Zone() == "" {
-		return errNoZone
+func placementKey(s services.ServiceID) string {
+	return fmt.Sprintf("%s/%s", placementPrefix, serviceKey(s))
```
xichen2020: Might be good to define constants for these formats in one place

kv/fake/fake_store.go
```diff
@@ -18,24 +18,25 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package kv
+package fake
```
cw9: will do the whole fakeStore move around thing in a separate diff

generated/proto/metadata/metadata.proto
```diff
@@ -0,0 +1,28 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+syntax = "proto3";
+
+package metadata;
+
+message metadata {
+  uint32 port = 1;
+  sfixed64 liveness_interval = 2;
```
cw9: lol, I don't know why this ends up as this type, yeah it prevents them from being varint encoded but we don't actually need that. changing them to int64

services/client/options.go
```diff
@@ -25,14 +25,28 @@ import (
 	"time"
 
 	"github.com/m3db/m3cluster/kv"
+	"github.com/m3db/m3cluster/services/heartbeat"
 	"github.com/m3db/m3x/instrument"
 )
 
-var errNoKVGen = errors.New("no KVGen function set")
+const (
+	defaultHBCheckInterval = 10 * time.Second
```
cw9: will do

services/client/options.go
```diff
@@ -60,20 +87,38 @@ type Options interface {
...
 func (o options) Validate() error {
 	if o.kvGen == nil {
 		return errNoKVGen
 	}
 
+	if o.hbGen == nil {
+		return errNoKVGen
+	}
+
+	if o.hbInterval == 0 {
```
cw9: sure will do

services/client/services.go
cw9: sg

services/client/services.go
cw9: sg

services/client/services.go
cw9: sg

services/client/util.go
```diff
@@ -194,3 +211,22 @@ func (su shardByIDAscending) Less(i, j int) bool {
 func (su shardByIDAscending) Swap(i, j int) {
 	su[i], su[j] = su[j], su[i]
 }
+
+func adKey(sid services.ServiceID, id string) string {
+	return fmt.Sprintf("[%s][%s]", serviceKey(sid), id)
```
xichen2020: Can we make these format consts? e.g.,
````
const (
    advertisementKeyFmt = "[%s][%s]"
    placementKeyFmt = "%s/%s"
    metadataKeyFmt = "%s/%s"
    serviceKeyFmt = "[%s]%s"
) 
```

Also out of curiosity, is there a reason why we are using different formats for different types of keys?

cw9: mostly to try to keep in sync with the old key format in statsdex, but actually "%s/%s" looks simpler and more clean now, unified to that format

services/client/util.go
```diff
@@ -32,6 +32,13 @@ import (
 	"github.com/m3db/m3cluster/shard"
 )
 
+const (
+	placementPrefix = "_sd.placement"
+	metadataPrefix  = "_sd.metadata"
+	keySperator     = "/"
```
xichen2020: Is this used anywhere?

services/client/util.go
xichen2020: Sounds good, does this mean we need to migrate existing keys then?

cw9: yeah we have to migrate everything since we are gonna use a different data store, but should be simple to use the existing m3admin code with new kv store

xichen2020: Sounds good then

cw9:
Sure, also depends on if we want to have the service instances send heartbeats to etcd directly or route through a sd service then to etcd store. But we surely can fall back to keeping records of the leases

PR #51, opened 2017-01-05
Add cache to kv client

cw9:
Use etcd revision instead of key version to compare values

@xichen2020 @robskillington

### DIFF
--- a/kv/etcd/etcd.go
+++ b/kv/etcd/etcd.go
@@ -45,14 +45,15 @@ func NewStore(c *clientv3.Client, opts Options) kv.Store {
 	scope := opts.InstrumentsOptions().MetricsScope()
 
 	store := &client{
-		opts:       opts,
-		kv:         c.KV,
-		watcher:    c.Watcher,
-		watchables: map[string]kv.ValueWatchable{},
-		retrier:    xretry.NewRetrier(opts.RetryOptions()),
-		logger:     opts.InstrumentsOptions().Logger(),
-		cacheFile:  opts.CacheFilePath(),
-		cache:      newCache(),
+		opts:           opts,
+		kv:             c.KV,
+		watcher:        c.Watcher,
+		watchables:     map[string]kv.ValueWatchable{},
+		retrier:        xretry.NewRetrier(opts.RetryOptions()),
+		logger:         opts.InstrumentsOptions().Logger(),
+		cacheFile:      opts.CacheFilePath(),
+		cache:          newCache(),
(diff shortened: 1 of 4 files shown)


kv/etcd/etcd.go
```diff
@@ -87,6 +102,10 @@ func (c *client) Get(key string) (kv.Value, error) {
 	r, err := c.kv.Get(ctx, c.opts.KeyFn()(key))
 	if err != nil {
 		c.m.etcdGetError.Inc(1)
+		cachedV, ok := c.getCache(key)
```
xichen2020: The expected behavior should be documented, i.e., we favor remote reads and only fall back to in-memory cache if the storage layer is unavailable

kv/etcd/etcd.go
```diff
@@ -269,6 +301,80 @@ func (c *client) CheckAndSet(key string, version int, v proto.Message) (int, err
...
+}
+
+func (c *client) mergeCache(key string, v *value) {
+	c.cache.Lock()
+	defer c.cache.Unlock()
+
+	cur, ok := c.cache.Values[key]
+	if !ok || v.isNewer(cur) {
+		c.cache.Values[key] = v
+		go c.syncFS()
```
xichen2020: This may cause a race w.r.t. writing out the cache file if we have two successive updates coming in since we are syncing to filesystem asynchronously, no? Probably better to have a long-lived goroutine watch for cache updates via a channel or something alike

kv/etcd/etcd.go
```diff
@@ -269,6 +301,80 @@ func (c *client) CheckAndSet(key string, version int, v proto.Message) (int, err
...
+	}
+}
+
+func (c *client) syncFS() error {
```
xichen2020: The returned error is silently ignored?

kv/etcd/etcd.go
cw9: will do

kv/etcd/etcd.go
xichen2020: In that case maybe don't return anything since you already logged and emit metric inside the function itself?

kv/etcd/etcd.go
cw9: yeah, good call, sounds good

kv/etcd/etcd.go
cw9: I feel it's nicer to return it in case in some cases the error matters

PR #52, opened 2017-01-05
Rename fake store to mem store

cw9:
@xichen2020 @robskillington


kv/mem/mem_store.go
```diff
@@ -18,74 +18,75 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package kv
+package mem
```
xichen2020: Could just call it `store.go` since it's already under the `mem` package

kv/mem/mem_store_test.go
```diff
@@ -18,33 +18,34 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package kv
+package mem
```
xichen2020: Could just call the file `store_test.go` for the same reason as above

kv/mem/mem_store.go
cw9: sounds good

PR #54, opened 2017-01-06
Return correct error

cw9:
@xichen2020 @robskillington


xichen2020:
Might be worth adding a regression test

cw9:
added test

PR #56, opened 2017-01-11
Add support for non sharded placement

cw9:
rename algo(_test).go to sharded(_test).go, helper(_test).go to sharded_helper(_test).go
added an algo to support non sharded placement in non_sharded.go
mainly trying to support the stateless backends in shunts, like m3dbingesters

There are 2 options I've been debating on to represent the non-sharded placement.
Both options will have rf set to 0 and every instance in the placement will only have necessary properties set but no shards.
option 1, set a numShards for the placement. This is for the case of m3db ingesters, if the number is set to 256, then only [0, 255) shards will be forwarded from shunts (test m3dbingesters). If set to 1024 then [0, 1023) shards will be forwarded (production m3dbingesters).
option 2, set no numShards for the placement as well. So the shunts need to config how many shards to send to which backend, but within the backend each instance are treated equal if the placement is non-sharded.

Decided to go with option 2, a non-sharded placement will have numShards = 0 and 0 shards on each instance, if shunt wants to dynamically config how much traffic goes to which backend, it needs to have a config on KV for that.

Now a Non-sharded placement requires numShards = 0, sharded = false. I'm allowing RF to be any number but I'm open to restrict that to 0 as well, don't feel strongly either way about this. 

@xichen2020 @robskillington @martin-mao


xichen2020:
I like option 2 better, stateless tiers shouldn't be aware of shards

services/types.go
```diff
@@ -197,6 +197,11 @@ type PlacementOptions interface {
 	// placed on instances other than the new instances in a replace operation
 	AllowPartialReplace() bool
 	SetAllowPartialReplace(allowPartialReplace bool) PlacementOptions
+
+	// Sharded describes whether a placement is sharded
+	// when set to false, no specific shards will be assigned to any instance
+	Sharded() bool
```
xichen2020: `IsSharded` might be better?

services/placement/service/placementservice.go
```diff
@@ -91,6 +84,33 @@ func (ps placementService) BuildInitialPlacement(
 	return p, ps.ss.SetIfNotExist(ps.service, p)
 }
 
+func (ps placementService) initPlacement(
+	instances []services.PlacementInstance,
+	ids []uint32,
+	rf int,
+) (services.ServicePlacement, error) {
+	if (rf == 0) == ps.opts.Sharded() {
```
xichen2020: This is a bit hard to read, can we spell out the two cases to make it a bit more readable?

services/placement/service/placementservice.go
```diff
@@ -91,6 +84,33 @@ func (ps placementService) BuildInitialPlacement(
...
+		return nil, fmt.Errorf("could not build initial placement with rf = %v and sharded = %v", rf, ps.opts.Sharded())
+	}
+
+	p, err := ps.algo.InitialPlacement(instances, ids)
```
xichen2020: what if `rf = 0` here? we still have one copy of the data?

services/placement/placement.go
```diff
@@ -80,12 +80,20 @@ func (p placement) NumShards() int {
 	return len(p.shards)
 }
 
+func (p placement) Sharded() bool {
+	return p.rf != 0
```
xichen2020: Using replica factor to indicate whether the placement is sharded is strange, in that non-sharded services essentially have one copy of the entire dataset. Can we add a boolean to indicate whether the placement is sharded or not?

services/placement/algo/non_sharded.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (a nonShardedAlgo) AddReplica(p services.ServicePlacement) (services.ServicePlacement, error) {
+	return nil, errAddReplicaOnNonSharded
+}
+
+func (a nonShardedAlgo) RemoveInstance(p services.ServicePlacement, instanceID string) (services.ServicePlacement, error) {
+	instances := p.Instances()
+	for i, instance := range instances {
+		if instance.ID() == instanceID {
+			instances = append(instances[:i], instances[i+1:]...)
```
xichen2020: can perhaps swap it with the last element and shrink the slice size to avoid a copy?

services/placement/service/placementservice.go
cw9: will probably not need this check once we added the isSharded field

services/placement/service/placementservice.go
cw9: the algo will be the non-sharded one when rf = 0, but I'm going to remove the relation between rf and IsSharded() as we discussed

services/placement/placement.go
cw9: sg, will change

services/types.go
cw9: sure, can do

services/placement/algo/non_sharded.go
cw9: will do

cw9:
> I like option 2 better, stateless tiers shouldn't be aware of shards
Sorry missed your comment, I agree, as I explained in the description, I'm afraid that it won't cover the case where we need to change the amount of traffic that goes to the stateless backend in shunts, which defeats the purpose of having the backend being dynamic. 
But at the same time, the amount of traffic that goes to a certain backend should be a config for shunts but not the backend itself. I'll probably move that control into a KV and leave the non-sharded placement really stateless.

services/placement/algo/sharded.go
```diff
@@ -0,0 +1,179 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
cw9: this file is moved from algo.go

xichen2020:
I think the value of having the backends dynamically discoverable is so that when new backends are added, we don't need to deploy the shunt tier. It just works.

PR #62, opened 2017-01-16
Add validation to config service options, add tag to metric scope

cw9:
@xichen2020 @robskillington

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -22,6 +22,8 @@ package etcd
 
 import (
 	"fmt"
+	"os"
+	"strings"
 	"sync"
 
 	"github.com/coreos/etcd/clientv3"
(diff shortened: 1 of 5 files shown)


client/etcd/client.go
```diff
@@ -46,17 +46,23 @@ const (
...
-func NewConfigServiceClient(opts Options) client.Client {
-	scope := opts.InstrumentOptions().MetricsScope()
+func NewConfigServiceClient(opts Options) (client.Client, error) {
+	if err := opts.Validate(); err != nil {
+		return nil, err
+	}
+
+	scope := opts.InstrumentOptions().
+		MetricsScope().
+		Tagged(map[string]string{"config_service_app": opts.AppID()})
```
xichen2020: maybe just call it `app_id` instead?

cw9: sure, can do

PR #63, opened 2017-01-17
Reuse lease in heartbeat store

cw9:
@xichen2020 @robskillington

### DIFF
--- a/services/client/services.go
+++ b/services/client/services.go
@@ -154,7 +154,7 @@ func (c *client) Advertise(ad services.Advertisement) error {
 			case <-ticker:
 				if isHealthy(ad) {
 					if err := hb.Heartbeat(serviceKey(sid), ad.InstanceID(), m.LivenessInterval()); err != nil {
-						c.logger.Infof("could not heartbeat service %s, %v", sid.String(), err)
+						c.logger.Errorf("could not heartbeat service %s, %v", sid.String(), err)
 						errCounter.Inc(1)
 					}
 				}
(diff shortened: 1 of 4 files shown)


services/heartbeat/etcd/store.go
```diff
@@ -64,16 +96,16 @@ func (c *client) Get(service string) ([]string, error) {
...
 }
 
-func heartbeatKey(service, id string) string {
-	return fmt.Sprintf(keyFormat, servicePrefix(service), id)
+func heartbeatKey(service, instance string) string {
+	return fmt.Sprintf(keyFormat, servicePrefix(service), instance)
 }
 
-func instanceID(key, service string) string {
+func instanceFromKey(key, service string) string {
```
xichen2020: maybe just `instanceKey` to be consistent with `heartbeatKey` and `leaseKey`?

services/client/services.go
```diff
@@ -143,12 +145,18 @@ func (c *client) Advertise(ad services.Advertisement) error {
...
+		errCounter := c.serviceTaggedScope(sid).Counter("heartbeat.error")
+
 		ticker := time.Tick(m.HeartbeatInterval())
 		for {
 			select {
 			case <-ticker:
 				if isHealthy(ad) {
-					hb.Heartbeat(serviceKey(ad.ServiceID()), ad.InstanceID(), m.LivenessInterval())
+					if err := hb.Heartbeat(serviceKey(sid), ad.InstanceID(), m.LivenessInterval()); err != nil {
+						c.logger.Infof("could not heartbeat service %s, %v", sid.String(), err)
```
xichen2020: Should this be an `Errorf`?

services/heartbeat/etcd/store.go
cw9: this function is to parse instance id from the heartbeat key

services/client/services.go
cw9: yeah good catch

services/heartbeat/etcd/store.go
xichen2020: Fair enough

PR #64, opened 2017-01-18
Remove placement validation in util function

cw9:
The converting function should not care about validation

@xichen2020 @robskillington


services/client/util.go
```diff
@@ -106,16 +106,11 @@ func PlacementFromProto(p placementproto.Placement) (services.ServicePlacement,
 		)
 	}
 
-	s := placement.NewPlacement().
+	return placement.NewPlacement().
```
xichen2020: Why do we remove validation?

cw9: we don't have to, but I feel it should be up to the placement algorithm and the placement service to decide whether the placement is valid rather than the util here, and that validation is done before all the persistence operation, if we do the validation here we can't even see what's in the placement if it's invalid.

xichen2020: fair enough

PR #66, opened 2017-01-21
Add Watch to heartbeat store and use in service discovery client

cw9:
@xichen2020 @robskillington

### DIFF
--- a/services/client/options.go
+++ b/services/client/options.go
@@ -30,16 +30,14 @@ import (
 )
 
 const (
-	defaultHeartbeatCheckInterval = 10 * time.Second
-	defaultInitTimeout            = 5 * time.Second
-	defaultGaugeInterval          = 10 * time.Second
+	defaultInitTimeout   = 15 * time.Second
+	defaultGaugeInterval = 10 * time.Second
 )
 
 var (
-	errNoKVGen                  = errors.New("no KVGen function set")
-	errNoHeartbeatGen           = errors.New("no HeartbeatGen function set")
-	errInvalidHeartbeatInterval = errors.New("non-positive interval for heartbeat check")
-	errInvalidInitTimeout       = errors.New("non-positive init timeout for service watch")
+	errNoKVGen            = errors.New("no KVGen function set")
+	errNoHeartbeatGen     = errors.New("no HeartbeatGen function set")
+	errInvalidInitTimeout = errors.New("non-positive init timeout for service watch")
 )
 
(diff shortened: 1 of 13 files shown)


services/heartbeat/etcd/store.go
```diff
@@ -101,6 +158,127 @@ func (c *client) Get(service string) ([]string, error) {
...
+	return true
+}
+
+func (c *client) processNotification(r clientv3.WatchResponse, w xwatch.Watchable, service string) {
+	err := r.Err()
+	if err != nil {
+		c.logger.Errorf("received error on watch channel: %v", err)
+		c.m.etcdWatchError.Inc(1)
+	}
+	// we need retry here because if Get() failed on an watch update,
```
xichen2020: feels like this function should return an error and it's the caller's responsibility to perform retries

services/client/options.go
```diff
@@ -30,16 +30,14 @@ import (
 )
 
 const (
-	defaultHeartbeatCheckInterval = 10 * time.Second
-	defaultInitTimeout            = 5 * time.Second
-	defaultGaugeInterval          = 10 * time.Second
+	defaultInitTimeout   = 15 * time.Second
```
cw9: because the Get() in kv store first calls etcd with a default timeout of 10 seconds, after the etcd call is timed out the kv store will fall back to the cache, so the initTimeout should be larger than 10 seconds here to really leverage the cache. I'll add a comment for it

services/client/options.go
```diff
@@ -57,13 +55,6 @@ type Options interface {
 	// SetInitTimeout sets the InitTimeout
 	SetInitTimeout(t time.Duration) Options
 
-	// HeartbeatCheckInterval is the interval for heartbeat check
```
cw9: because previously I call heartbeatStore.Get() every HeartbeatCheckInterval, now if waits for the notification on the watch to call Get()

services/heartbeat/etcd/store.go
cw9: the whole function is actually only called in one place, I extracted it out here to make the original loop shorter, maybe it's better to move the code there and retry there?

services/client/options.go
xichen2020: Right yeah makes sense

services/heartbeat/etcd/store.go
xichen2020: Yeah that'll be better, doesn't feel intuitive to retry in a `process...` method

PR #71, opened 2017-01-24
Drop Go1.6 Travis builds

prateek:
/cc @cw9 @robskillington @xichen2020

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: go
 go:
-  - 1.6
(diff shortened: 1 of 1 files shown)


cw9:
remove the glide version pinning in this file as well?

prateek:
@cw9, nah. still waiting to see how the upstream glide issue is fixed. `m3db` is still doing the pinning too. Will do both once glide's issue is fixed.

PR #76, opened 2017-01-30
Make kv.Watch non-blocking

cw9:
@xichen2020 @robskillington

### DIFF
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -143,27 +143,55 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
 	c.Lock()
 	watchable, ok := c.watchables[key]
 	if !ok {
-		watchChan := c.watchChan(key)
-		c.m.etcdWatchCreate.Inc(1)
-
 		watchable = kv.NewValueWatchable()
 		c.watchables[key] = watchable
 
 		go func() {
 			ticker := time.Tick(c.opts.WatchChanCheckInterval())
 
+			var (
+				watchChan clientv3.WatchChan
+				err       error
+			)
 			for {
+				if watchChan == nil {
+					c.m.etcdWatchCreate.Inc(1)
+					watchChan, err = c.watchChanWithTimeout(key)
+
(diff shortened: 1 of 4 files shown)


kv/etcd/store.go
```diff
@@ -220,27 +261,22 @@ func (c *client) tryCleanUp(key string) bool {
...
-func (c *client) update(w kv.ValueWatchable, key string) error {
-	newValue, err := c.Get(key)
-	if err == kv.ErrNotFound {
-		// nothing to update
-		return nil
+	if execErr := c.retrier.Attempt(func() error {
+		newValue, err = c.Get(key)
+		if err == kv.ErrNotFound {
+			// do not retry on ErrNotFound
+			return nil
```
xichen2020: Should we wrap this as a non-retryable error instead of just returning nil here?

kv/etcd/store.go
```diff
@@ -188,16 +216,29 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
...
-		// receive initial notification once the watch channel is created
-		clientv3.WithCreatedNotify(),
-	)
+func (c *client) watchChanWithTimeout(key string) (clientv3.WatchChan, error) {
+	doneCh := make(chan struct{})
+
+	var watchChan clientv3.WatchChan
+	go func() {
+		watchChan = c.watcher.Watch(
+			context.Background(),
```
xichen2020: Can we directly pass in a context with timeout here instead of doing a select down below?

Also if it does time out, is there a way to clean up the watcher?

cw9: we can pass in a context here, but that means the watch will be ended every context.Timeout, I feel that's unnecessary when the watch in stable.

Yeah the watchChan is essentially a receive channel, there is no good way to clean it up.

kv/etcd/store.go
cw9: Good call, will fix, I reopened the pr in https://github.com/m3db/m3cluster/pull/77 btw

PR #77, opened 2017-01-30
Make watch() non-blocking

cw9:
This diff added timeout and retry logic around etcd watch creation to prevent kv.Watch() from hanging due to etcd watch creation not responding.

There should be no usage change for kv.Watch() users. The kv.Watch() call is guaranteed to be non-blocking and the watch itself manages all the error cases so the users of kv.Watch() don't have to.

@xichen2020 @robskillington

### DIFF
--- a/etcd/watch_helper/types.go
+++ b/etcd/watch_helper/types.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package watchhelper
+
+import (
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/m3db/m3x/instrument"
+)
+
+// WatchHelper helps maintaining etcd watch on a key
+type WatchHelper interface {
+	// Run watches the key forever until the CheckAndStopFn returns true
+	Run(key string)
+}
+
+// UpdateFn is called when an event on the watch channel happens
+type UpdateFn func(key string) error
+
+// CheckAndStopFn is called every once a while
+// to check and stop the watch if needed
+type CheckAndStopFn func(key string) bool
+
+// Options are options for the etcd watch helper
+type Options interface {
+	// Watcher is the etcd watcher
+	Watcher() clientv3.Watcher
+	// SetWatcher sets the Watcher
+	SetWatcher(w clientv3.Watcher) Options
+
+	// UpdateFn is the function called when a notification on a key is received
+	UpdateFn() UpdateFn
+	// SetUpdateFn sets the UpdateFn
+	SetUpdateFn(f UpdateFn) Options
+
+	// CheckAndStopFn is the function called periodicaly to check if a watch should be stopped
+	CheckAndStopFn() CheckAndStopFn
+	// SetCheckAndStopFn sets the CheckAndStopFn
+	SetCheckAndStopFn(f CheckAndStopFn) Options
+
+	// WatchOptions is a set of options for the etcd watch
+	WatchOptions() []clientv3.OpOption
+	// SetWatchOptions sets the WatchOptions
+	SetWatchOptions(opts []clientv3.OpOption) Options
+
+	// WatchChanCheckInterval will be used to periodicaly check if a watch chan
+	// is no longer being subscribed and should be closed
+	WatchChanCheckInterval() time.Duration
+	// SetWatchChanCheckInterval sets the WatchChanCheckInterval
+	SetWatchChanCheckInterval(t time.Duration) Options
+
+	// WatchChanResetInterval is the delay before resetting the etcd watch chan
+	WatchChanResetInterval() time.Duration
+	// SetWatchChanResetInterval sets the WatchChanResetInterval
+	SetWatchChanResetInterval(t time.Duration) Options
+
+	// WatchChanInitTimeout is the timeout for a watchChan initialization
+	WatchChanInitTimeout() time.Duration
+	// SetWatchChanInitTimeout sets the WatchChanInitTimeout
+	SetWatchChanInitTimeout(t time.Duration) Options
+
+	// InstrumentsOptions is the instrument options
+	InstrumentsOptions() instrument.Options
+	// SetInstrumentsOptions sets the InstrumentsOptions
+	SetInstrumentsOptions(iopts instrument.Options) Options
+
+	// Validate validates the options
+	Validate() error
+}
--- a/etcd/watch_helper/watch_helper.go
+++ b/etcd/watch_helper/watch_helper.go
@@ -1,0 +1,148 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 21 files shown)


xichen2020:
Would be nice if you could include a brief description of the PR in the summary going forward (e.g., what it does, why it's necessary, any gotchas etc), which would be super helpful to reviewers

etcd/watch_helper/types.go
```diff
@@ -0,0 +1,88 @@
...
+	// SetWatcher sets the Watcher
+	SetWatcher(w clientv3.Watcher) Options
+
+	// UpdateFn is the function called when a notification on a key is received
+	UpdateFn() UpdateFn
+	// SetUpdateFn sets the UpdateFn
+	SetUpdateFn(f UpdateFn) Options
+
+	// CheckAndStopFn is the function called periodicaly to check if a watch should be stopped
+	CheckAndStopFn() CheckAndStopFn
```
xichen2020: `TickFn` might be a better name?

services/heartbeat/etcd/store.go
```diff
@@ -142,109 +168,93 @@ func (c *client) Heartbeat(service, instance string, ttl time.Duration) error {
...
-		servicePrefix(service),
-		// WithPrefix so that the watch will receive any changes
-		// from the instances under the service
-		clientv3.WithPrefix(),
-		// periodically (appx every 10 mins) checks for the latest data
-		// with or without any update notification
-		clientv3.WithProgressNotify(),
-		// receive initial notification once the watch channel is created
-		clientv3.WithCreatedNotify(),
+func (c *client) update(key string) error {
```
xichen2020: The logic here looks similar to the kv etc store, any chance we could move the `update` and `checkAndStop` to a base `client` struct along with the corresponding data structures and embed it in the kv etc store and heartbeat etcd store?

testutil/blackhole_watcher.go
```diff
@@ -0,0 +1,48 @@
+package testutil
```
xichen2020: `mocks` might be a better name?

testutil/blackhole_watcher.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"golang.org/x/net/context"
+)
+
+// MockWatcher mocks an etcd client that just blackholes a few watch requests
+type MockWatcher struct {
```
xichen2020: If the package is called `mocks`, this could just be `Watcher` and the filename is `watcher.go`

testutil/blackhole_watcher.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Watch is implementing etcd clientv3 Watcher interface
+func (m *MockWatcher) Watch(ctx context.Context, key string, opts ...clientv3.OpOption) clientv3.WatchChan {
+	m.Lock()
+
+	if m.failed < m.failTotal {
+		m.failed++
+		m.Unlock()
+
+		time.Sleep(time.Minute)
```
xichen2020: hm why do we need to sleep here?

etcd/watch_helper/types.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/m3db/m3x/instrument"
+)
+
+// WatchHelper helps maintaining etcd watch on a key
+type WatchHelper interface {
```
cw9: will do WatchManager

etcd/watch_helper/types.go
```diff
@@ -0,0 +1,88 @@
...
+	// Run watches the key forever until the CheckAndStopFn returns true
+	Run(key string)
```
cw9: can do

etcd/watch_helper/types.go
cw9: I'm ok with either way, but feels TickFn is missing the idea of "Stop" in it.

etcd/watch_helper/watch_helper.go
```diff
@@ -0,0 +1,148 @@
...
+	scope := opts.InstrumentsOptions().MetricsScope()
+	return &watchHelper{
+		opts:   opts,
+		logger: opts.InstrumentsOptions().Logger(),
+		m: clientMetrics{
+			etcdWatchCreate: scope.Counter("etcd-watch-create"),
+			etcdWatchError:  scope.Counter("etcd-watch-error"),
+			etcdWatchReset:  scope.Counter("etcd-watch-reset"),
+		},
+
```
cw9: will do

etcd/watch_helper/watch_helper.go
```diff
@@ -0,0 +1,148 @@
...
+		checkAndStopFn: opts.CheckAndStopFn(),
+	}, nil
+}
+
+type watchHelper struct {
+	opts   Options
+	logger xlog.Logger
+	m      clientMetrics
+
+	updateFn       func(key string) error
```
cw9: sure can do

etcd/watch_helper/watch_helper.go
```diff
@@ -0,0 +1,148 @@
...
+	checkAndStopFn func(key string) bool
```
cw9: sure

services/heartbeat/etcd/store.go
cw9: yeah... since Golang does not have generic, this very similar code has to be duplicated, because one store is managing kv.ValueWatchable and the other one is managing watch.Watchable. Also the logic is slightly different. the kv store cares about versioning and the other one don't, thus the separation

testutil/blackhole_watcher.go
cw9: will do

testutil/blackhole_watcher.go
cw9: sg

testutil/blackhole_watcher.go
cw9: to mimic a blackhole that does not respond for a long time.

xichen2020: In that case perhaps the mock watcher ctor should probably take in a `OnFail` method so you can control its behavior more explicitly

robskillington:
@cw9 can you perhaps give a short description of why we are making this change and what consumers will need to change their behavior and how?

etcd/watchmanager/manager.go
```diff
@@ -0,0 +1,148 @@
...
+			w.opts.WatchOptions()...,
+		)
+		close(doneCh)
+	}()
+
+	select {
+	case <-doneCh:
+		return watchChan, nil
+	case <-time.After(w.opts.WatchChanInitTimeout()):
+		return nil, fmt.Errorf("etcd watch chan creation timed-out on key: %s", key)
```
robskillington: Perhaps also include the timeout duration in the error message?
```go
timeout := w.opts.WatchChanInitTimeout()
fmt.Errorf("etcd watch create timed out after %s for key: %s", timeout.String(), key)
```

etcd/watchmanager/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	var (
+		watchChan clientv3.WatchChan
+		err       error
+	)
+	for {
+		if watchChan == nil {
+			w.m.etcdWatchCreate.Inc(1)
+			watchChan, err = w.watchChanWithTimeout(key)
+			if err != nil {
+				w.logger.Errorf("could not create etcd watch, %v", err)
```
robskillington: Perhaps use a `:` to list the error here like other log messages:
```go
w.logger.Errorf("could not create etcd watch: %v", err)
```

etcd/watchmanager/manager.go
```diff
@@ -0,0 +1,148 @@
...
+
+				// NB(cw) when we failed to create a etcd watch channel
+				// we do a get for now and will try to recreate the watch chan later
+				if err = w.updateFn(key); err != nil {
```
robskillington: Why is it worth doing a get if watch fails?  If we need a watch shouldn't we just continue to acquire it?

etcd/watchmanager/manager.go
```diff
@@ -0,0 +1,148 @@
...
+				w.m.etcdWatchError.Inc(1)
+				// do not stop here, even though the update contains an error
+				// we still take this chance to attemp a Get() for the latest value
+			}
+
+			if err = w.updateFn(key); err != nil {
+				w.logger.Errorf("received notification for key %s, but failed to get value: %v", key, err)
+			}
+		case <-ticker:
+			if w.tickAndStopFn(key) {
```
robskillington: Instead of reading a ticker and check if it should stop, wouldn't it be better to immediately signal stop through some channel?  Otherwise we're going to be waking up goroutines that don't necessarily need to be woken.

etcd/watchmanager/manager.go
cw9: sure, will do

etcd/watchmanager/manager.go
cw9: sg

etcd/watchmanager/manager.go
cw9: So if we can never get a watch established, we can still (regularly) try to get the latest value and push to the users rather than sorely depend on the etcd watch for notifications

etcd/watchmanager/manager.go
cw9: The stop here is mostly triggered by closing all the watches on a watchable, so calling one watch.Close() does not necessarily trigger the close on this go routine, I could replace the tick here with a wait on done channel, but then the users of this watch manager (the kv watch or the heartbeat store watch) has to handle the logic of regularly check if all the watches on a certain watchable is closed and send out the close signal to the watch manager here. So those users have to duplicate the logic and have another go routine to do the check.

cw9:
Updated the diff description

cw9:
bump @xichen2020 @robskillington

xichen2020:
I still have the outstanding comment on passing the `OnFail` method to the mock watcher ctor, also you may need to rebase due to the conflicts

PR #78, opened 2017-02-02
Toggle to control set in BuildInitialPlacement

prateek:
Need control over this to reset placements in `m3em`


prateek:
spoke to @cw9, we're going to instead add the ability to delete existing placements

Issue #81, opened 2017-02-03
Try returning Initializing shards back to original owner


cw9:
There might be a case where a node was recently added and got assigned shards in Initializing state with sourceID to where the shard is originally from, if this node is removed before finish Initializing all of its shards, those shards will be redistributed to the cluster with no preference to try returning it back to the original node (the ownership may go back to the original node or it will end up in another node). 

Returning the shard back to it's original owner should reduce some amount of bootstrapping work and it should be safe because the original node still has the shard in leaving state and been taking all the writes while the new node is bootstrapping. 

But the original node is not guaranteed to be able to take the shard back, because while it's holding the shard in Leaving state, if another placement operation happened and moved another replica of that shard to the same rack of the original node, this shard will not be able to be moved back due to rack check.

PR #82, opened 2017-02-06
Prefer to return the Initializing shards back to source instance

cw9:
Addressing: https://github.com/m3db/m3cluster/issues/81

@xichen2020 @robskillington

### DIFF
--- a/services/placement/algo/sharded_helper.go
+++ b/services/placement/algo/sharded_helper.go
@@ -62,7 +62,7 @@ type placementHelper struct {
 	totalWeight        uint32
 	rf                 int
 	uniqueShards       []uint32
-	instances          []services.PlacementInstance
+	instances          map[string]services.PlacementInstance
 	opts               services.PlacementOptions
 }
 
@@ -116,7 +116,7 @@ func (ph *placementHelper) MoveShard(candidateShard shard.Shard, from, to servic
 	return true
 }
 
-func (ph placementHelper) HasNoRackConflict(shard uint32, from services.PlacementInstance, toRack string) bool {
+func (ph *placementHelper) HasNoRackConflict(shard uint32, from services.PlacementInstance, toRack string) bool {
 	if from != nil {
 		if from.Rack() == toRack {
 			return true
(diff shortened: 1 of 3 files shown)


services/placement/algo/sharded_helper.go
```diff
@@ -142,21 +142,27 @@ func (ph *placementHelper) GeneratePlacement() services.ServicePlacement {
 		}
 	}
 	return placement.NewPlacement().
-		SetInstances(ph.instances).
+		SetInstances(ph.getInstances()).
```
xichen2020: Can we come up with a better name than `get..`? Perhaps `instanceList` or something along those lines?

services/placement/algo/sharded_helper.go
```diff
@@ -142,21 +142,27 @@ func (ph *placementHelper) GeneratePlacement() services.ServicePlacement {
...
 }
 
-func (ph placementHelper) PlaceShards(shards []shard.Shard, from services.PlacementInstance) error {
+func (ph *placementHelper) PlaceShards(shards []shard.Shard, from services.PlacementInstance) error {
 	shardSet := getShardMap(shards)
 	if from != nil {
+		// NB(cw) when removing an adding instance that has not finished bootstrapping its
+		// Initializing shards, prefer to return those Initializing shards back to the leaving instance
+		// to reduce some bootstrapping work in the cluster.
+		if err := ph.returnInitializingShardsToSource(shardSet, from); err != nil {
```
xichen2020: Will this method modify `shardSet` or `from` in any way? If so, would it be okay if it returns an error and `shardSet` and `from` has been changed?

cw9: yeah it will modify, shardSet is an internal var in this function, so I'm not worried about it. The changes made to the "from" instance before error would concern me more, but since all the usages of the PlaceShards() has to return on error and we have cloned the placement before every time we create the placement helper, I think this is ok.

xichen2020: Okay fair enough

PR #83, opened 2017-02-08
Add History function to kv API

cw9:
Add a History function to get values of past versions

@xichen2020 @robskillington

### DIFF
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -22,6 +22,7 @@ package etcd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"sync"
@@ -38,7 +39,11 @@ import (
 
 const etcdVersionZero = 0
 
-var noopCancel func()
+var (
+	noopCancel func()
+
+	errInvalidHistoryVersion = errors.New("invalid version range")
+)
 
 // NewStore creates a kv store based on etcd
 func NewStore(c *clientv3.Client, opts Options) (kv.Store, error) {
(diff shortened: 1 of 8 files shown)


kv/etcd/store.go
```diff
@@ -150,17 +155,72 @@ func (c *client) get(key string) (kv.Value, error) {
...
+	if r.Count == 0 {
+		return nil, kv.ErrNotFound
+	}
+
+	numValue := to - from
+
+	latestKV := r.Kvs[0]
+	latestVersion := int(latestKV.Version)
+	latestModRev := latestKV.ModRevision
+	createRev := latestKV.CreateRevision
```
xichen2020: What's create revision?

kv/etcd/store.go
```diff
@@ -150,17 +155,72 @@ func (c *client) get(key string) (kv.Value, error) {
...
+
+	if latestVersion < from {
+		return nil, errors.New("no value available in the version range")
```
xichen2020: Maybe just return `nil, nil` here?

kv/etcd/store.go
```diff
@@ -150,17 +155,72 @@ func (c *client) get(key string) (kv.Value, error) {
...
-
 	v := newValue(r.Kvs[0].Value, r.Kvs[0].Version, r.Kvs[0].ModRevision)
 
 	c.mergeCache(key, v)
 
 	return v, nil
 }
 
+func (c *client) History(key string, from, to int) ([]kv.Value, error) {
+	if from >= to || from < 0 || to < 0 {
```
cw9: sure, can do

kv/etcd/store.go
```diff
@@ -150,17 +155,72 @@ func (c *client) get(key string) (kv.Value, error) {
...
+		return nil, errInvalidHistoryVersion
+	}
+
+	newKey := c.opts.KeyFn()(key)
+
+	ctx, cancel := c.context()
+	r, err := c.kv.Get(ctx, newKey)
+	cancel()
```
cw9: sure

kv/etcd/store.go
cw9: The store revision when this key is created

kv/etcd/store.go
cw9: sg

kv/types.go
```diff
@@ -95,4 +95,7 @@ type Store interface {
 	// CheckAndSet stores the value for the given key if the current version
 	// matches the provided version
 	CheckAndSet(key string, version int, v proto.Message) (int, error)
+
+	// History returns the value for a key in version range [from, to)
+	History(key string, from, to int) ([]Value, error)
```
cw9: GetBetween seems too vague, GetRange might sound like a function that lists all the keys in a range (which will be added soon)

kv/etcd/store.go
cw9: yeah, I guess I don't really need to check this, will remove

kv/etcd/store.go
```diff
@@ -163,15 +163,16 @@ func (c *client) get(key string) (kv.Value, error) {
 }
 
 func (c *client) History(key string, from, to int) ([]kv.Value, error) {
-	if from >= to || from < 0 || to < 0 {
+	if from > to || from < 0 || to < 0 {
 		return nil, errInvalidHistoryVersion
 	}
```
xichen2020: As a simple optimization, add `if from == to { return nil, nil }` to avoid unnecessary work

PR #84, opened 2017-02-14
Add Delete to kv store

cw9:
Add a delete function to kv interface and placement service interface.

Now the watch returned from kv.Watch() may get nil values even after the watch has received initial notification because it could be triggered by the deletion of the key.  So the user of that watch needs to do nil check on watch.Get()

The sd.Watch() will ignore the deletion update in runtime and fall back to the last known valid topology, so the topology returned from that watch is always valid. But once the sd client restarts, it will won't fall back to that cached value because it will get the nil value and ErrNotFound from the data store, so it will block/time out on initialization.


kv/etcd/store.go
```diff
@@ -276,9 +277,21 @@ func (c *client) update(key string) error {
 
 	curValue := w.Get()
 	if curValue == nil {
```
prateek: could this be simplified to the following:

```
if currValue == nil || newValue == nil {
    return w.Update(newValue)
}
```

kv/etcd/store_test.go
```diff
@@ -521,7 +521,106 @@ func TestHistory(t *testing.T) {
...
+	require.Equal(t, 2, version)
+
+	v, err = store.Get("foo")
+	require.NoError(t, err)
+	verifyValue(t, v, "bar2", 2)
+
+	v, err = store.Delete("foo")
+	require.NoError(t, err)
+	verifyValue(t, v, "bar2", 2)
+
```
prateek: Could you add a check that a `Delete` after a `Delete` fails.

prateek:
I don't understand the implication for users of the API if `sd.Watch()` ignores deletes. Why ignore instead of passing that information on?

kv/etcd/store.go
cw9: I can do 
```
if curValue == nil && newValue == nil {
    return nil
}
if curValue == nil || newValue == nil {
    return w.Update(newValue)
}
```

kv/etcd/store_test.go
cw9: sure, will do

cw9:
> I don't understand the implication for users of the API if sd.Watch() ignores deletes. Why ignore instead of passing that information on?

So previously when there are invalid topology updates, we ignore them in sd.Watch(), this is to prevent human mistakes from spreading across the cluster, I feel delete should probably be handled the same way. Also unlike the kv.Watch() call, we block on the initialization of sd.Watch() to wait for a valid topology to be available. So I feel the users of this API expect the topologies passed from sd.Watch() to be valid.

kv/etcd/store.go
prateek: ah so can't update until you receive the first value?

prateek:
That makes sense. What happens if a client is watching on a topology, it gets deleted, and then a new topology is created for the same ID - does the client watching get notified of the update?

cw9:
> That makes sense. What happens if a client is watching on a topology, it gets deleted, and then a new topology is created for the same ID - does the client watching get notified of the update?

Yes it will get notified, have it covered in unit test

kv/etcd/store.go
cw9: yep

prateek: cool, makes sense.

PR #85, opened 2017-02-16
Change Unadvertise to clean up the existing heartbeat for the instance

cw9:
Previously the Unadvertise call will only stop the go routine that keeps heartbeating the instance, now it also deletes the heartbeat in the data store.

### DIFF
--- a/services/client/services.go
+++ b/services/client/services.go
@@ -170,17 +170,23 @@ func (c *client) Unadvertise(sid services.ServiceID, id string) error {
 		return err
 	}
 
-	c.Lock()
-	defer c.Unlock()
-
 	key := adKey(sid, id)
+
+	c.Lock()
 	ch, ok := c.adDoneChs[key]
-	if !ok {
-		return fmt.Errorf("service %s, instance %s is not being advertised", sid.String(), id)
+	if ok {
+		// if this client is advertising the instance, stop it
+		close(ch)
+		delete(c.adDoneChs, key)
 	}
-	ch <- struct{}{}
-	delete(c.adDoneChs, key)
-	return nil
+	c.Unlock()
+
+	hbStore, err := c.getHeartbeatStore(sid.Zone())
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


services/client/services.go
```diff
@@ -170,17 +170,23 @@ func (c *client) Unadvertise(sid services.ServiceID, id string) error {
...
-	c.Lock()
-	defer c.Unlock()
-
 	key := adKey(sid, id)
+
+	c.Lock()
 	ch, ok := c.adDoneChs[key]
-	if !ok {
-		return fmt.Errorf("service %s, instance %s is not being advertised", sid.String(), id)
+	if ok {
```
xichen2020: nit: if `ch, ok := c.adDoneChs[key]; ok {...}`

services/heartbeat/etcd/store.go
```diff
@@ -294,6 +306,47 @@ func servicePrefix(service string) string {
...
+
+type leaseCache struct {
+	sync.RWMutex
+
+	leases map[string]map[time.Duration]clientv3.LeaseID
+}
+
+func (c *leaseCache) get(service, instance string, ttl time.Duration) (clientv3.LeaseID, bool) {
+	c.RLock()
+	leases, ok := c.leases[heartbeatKey(service, instance)]
```
xichen2020: Might be better to just defer `c.RUnlock()`  so you don't need to track where the function might return

services/heartbeat/etcd/store.go
```diff
@@ -294,6 +306,47 @@ func servicePrefix(service string) string {
...
+	id, ok := leases[ttl]
+	c.RUnlock()
+
+	return id, ok
+}
+
+func (c *leaseCache) put(service, instance string, ttl time.Duration, id clientv3.LeaseID) {
+	key := heartbeatKey(service, instance)
+
+	c.Lock()
```
xichen2020: Same, `defer c.Lock()` might be better

cw9:
Addressed all comments

PR #86, opened 2017-02-16
Allow adding leaving instance back to placement

cw9:
Previously when adding an instance, we bail if the adding instance is already in placement,
this diff adds an exception if the existing instance is a leaving instance, meaning it has only Leaving shards.

This supports backing out of an instance removal operation.

The algo will first try taking those leaving shards back, and then do a normal addInstance if necessary to even the load.

@robskillington @xichen2020

### DIFF
--- a/services/placement/algo/sharded.go
+++ b/services/placement/algo/sharded.go
@@ -162,7 +162,8 @@ func (a rackAwarePlacementAlgorithm) addInstance(p services.ServicePlacement, ad
 		if !placement.IsInstanceLeaving(instance) {
 			return nil, errAddingInstanceAlreadyExist
 		}
-		p, addingInstance = a.moveLeavingShardsBack(p, instance)
+		p = a.moveLeavingShardsBack(p, instance)
+		addingInstance = instance
 	}
 
 	ph := newAddInstanceHelper(p, addingInstance, a.opts)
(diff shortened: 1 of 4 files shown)


services/placement/algo/sharded.go
```diff
@@ -157,10 +158,13 @@ func (a rackAwarePlacementAlgorithm) ReplaceInstance(
 }
 
 func (a rackAwarePlacementAlgorithm) addInstance(p services.ServicePlacement, addingInstance services.PlacementInstance) (services.ServicePlacement, error) {
-	if _, exist := p.Instance(addingInstance.ID()); exist {
-		return nil, errAddingInstanceAlreadyExist
-
+	if instance, exist := p.Instance(addingInstance.ID()); exist {
```
xichen2020: Can we do the following instead so `movingLeavingShardsBack` doesn't need to return the instance?
```
if _, exists := p.Instance(addingInstance.ID()); exist {
      if !placement.IsInstanceLeaving(addingInstance) {
          return nil, errAddingInstanceAlreadyExist
      }
      p = a.moveLeavingShardsBack(p, addingInstance)
}
```

cw9: Sometimes when the users tries to add an instance, the instance will have correct rack,zone,weight info but not neccessarily the correct shards info, using the instance found from the placement gets the correct shards data

xichen2020: Ok in that case maybe
```
var exists bool
if addingInstance, exists = p.Instance(addingInstance.ID()); exist {
      if !placement.IsInstanceLeaving(addingInstance) {
          return nil, errAddingInstanceAlreadyExist
      }
      p = a.moveLeavingShardsBack(p, addingInstance)
}
```

PR #88, opened 2017-02-20
Change KeyFn() to ApplyPrefix() in etcd kv options

cw9:
Remove the KeyFn() in the kv options and add ApplyPrefix() to prevent
users from adding suffix to the key, which might cause issues in etcd
usage in the future

For example: etcd allows adding a WithPrefix() in the Get calls, but if
the key we passed in was wrapped with a suffix, the result might be
confusing.

@xichen2020 @robskillington @schallert @prateek

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -43,7 +43,7 @@ import (
 const (
 	keySeparator    = "/"
 	cacheFileFormat = "%s-%s.json"
-	kvPrefix        = "_kv"
+	kvPrefix        = "_kv/"
 )
 
 type newClientFn func(endpoints []string) (*clientv3.Client, error)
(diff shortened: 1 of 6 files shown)


services/client/services_test.go
```diff
@@ -701,9 +700,7 @@ func testSetup(t *testing.T) (Options, func(), *mockHBGen) {
 			ec,
 			etcdKV.NewOptions().
 				SetWatchChanCheckInterval(100*time.Millisecond).
-				SetKeyFn(func(key string) string {
-					return fmt.Sprintf("[%s][%s]", zone, key)
-				}),
+				SetPrefix(zone+"/"),
```
prateek: nit: could you use `fmt.Sprintf` here

client/etcd/client.go
```diff
@@ -204,12 +200,23 @@ func cacheFileForZone(cacheDir, appID, zone string) string {
...
 }
 
 func fileName(appID, zone string) string {
 	cacheFileName := fmt.Sprintf(cacheFileFormat, appID, zone)
 
 	return strings.Replace(cacheFileName, string(os.PathSeparator), "_", -1)
 }
+
+func prefix(env string) string {
+	res := concat(kvPrefix, keySeparator)
```
prateek: nit: could you make this a constant too?

PR #89, opened 2017-02-21
[heartbeat] Store instance proto data in hb store

schallert:
As we move to a model where the Heartbeat can be used for selecting
placement candidates from a standby pool in addition to just maintaining
IDs of alive instances, we need to store a bit more state in the
heartbeat store along with just instance ID. This modifies the
`heartbeat.Store` interface to accept a `PlacementInstance` object
describing the instance that is heartbeating and store the
protobuf-serialized version of the instance in the heartbeat store.

When querying the Heartbeat store, the deserialized `PlacementInstance`s
will be returned.

RFR @cw9


services/client/services_test.go
```diff
@@ -105,7 +111,10 @@ func TestAdvertiseErrors(t *testing.T) {
 	require.Error(t, err)
 	require.Equal(t, errNoInstanceID, err)
 
-	ad = ad.SetInstanceID("i1")
+	pi := placement.NewInstance()
```
cw9: nit: pi -> i1? and move the SetID() up to this line?

services/heartbeat/types.go
```diff
@@ -23,16 +23,17 @@ package heartbeat
...
 
 // Store is a heartbeat store
 type Store interface {
 	// Heartbeat sends heartbeat for a service instance with a ttl
-	Heartbeat(service, instance string, ttl time.Duration) error
+	Heartbeat(service string, instance services.PlacementInstance, ttl time.Duration) error
 
 	// Get gets healthy instances for a service
-	Get(service string) ([]string, error)
+	Get(service string) ([]services.PlacementInstance, error)
```
cw9: I'd consider having 2 different Get(), one gets instance info, the other one gets the ids only (could add WithKeysOnly() in etcd calls) so that use case could be cheaper, also do you think getting keys only is enough for the watch? cause the watch in services interface only cares the instance ids from heartbeat store.

schallert: I like the idea of having a more lightweight `Get` function that only returns the IDs. As for the watch question, I've been going back and forth on this while trying to predict the various use cases. I think what I'm going to do is keep the `heartbeat.Store.Watch` function as returning just the string IDs, but add a `WatchInstances` function that will return a watch that deserializes the values. Then I'll add a `WithInstances(bool)` function or similar to the `services.QueryOptions` interface, and that will determine which of the two `store.Watch*` functions is used when the caller calls `services.Services.Watch(...)`

schallert:
I just added a commit that makes a few more API changes to the heartbeat store. Specifically the `HeartbeatStore` functions now take a full `ServiceID` object rather than just a string, since we need both environment and service name to construct the heartbeat key.

services/client/util.go
```diff
@@ -53,171 +49,6 @@ func metadataToProto(m services.Metadata) metadataproto.Metadata {
 	}
 }
```
cw9: let's move these meta proto util functions out as well

services/heartbeat/etcd/store.go
```diff
@@ -302,8 +350,13 @@ func instanceFromKey(key, servicePrefix string) string {
 	)
 }
 
-func servicePrefix(service string) string {
-	return fmt.Sprintf(keyFormat, heartbeatKeyPrefix, service)
+// heartbeats for a service "svc" in env "test" should be stored under
+// "_hb/test/svc"
```
cw9: nit: just a question to throw out there: if the env is empty, should we do _hb/svc or _hb//svc?

schallert:
@cw9 for the env empty prefix case I went with including a `"default_env"` environment. I was worried that if we were to just do `_hb/svc` then this would prevent use cases such as iterating over the `_hb` keys to enumerate all environments which currently have instances heartbeating, since direct subkeys of `_hb/` would be a mix of environment types and services types.

services/client/services.go
```diff
@@ -299,6 +306,14 @@ func (c *client) Watch(sid services.ServiceID, opts services.QueryOptions) (xwat
 	return w, err
 }
 
+func (c *client) HeartbeatService(zone string) (services.HeartbeatService, error) {
+	if zone == "" {
```
cw9: nit: this case is probably fine, if getHeartbeatService() can create a valid store for zone "", then we should just return it

services/types.go
```diff
@@ -50,6 +50,9 @@ type Services interface {
 
 	// PlacementService returns a client of Placement Service
 	PlacementService(service ServiceID, popts PlacementOptions) (PlacementService, error)
+
+	// HeartbeatService returns a heartbeat store for the given zone.
+	HeartbeatService(zone string) (HeartbeatService, error)
```
cw9: hmm... I actually prefer not to expose a new argument of "zone" here, since all the other methods in this interface pretty much all take in a service id, maybe just do that here as well?

cw9:
On the defaultEnv topic, I checked the way how placement treats the emtpy environment now, it actually just leave it empty when there is no env, perhaps just keep the behavior the same for now until there is really an issue with it?

services/client/services.go
```diff
@@ -332,20 +332,22 @@ func (c *client) getPlacementValue(sid services.ServiceID) (kv.Value, error) {
 	return v, nil
 }
 
-func (c *client) getHeartbeatService(zone string) (services.HeartbeatService, error) {
+func (c *client) getHeartbeatService(sid services.ServiceID) (services.HeartbeatService, error) {
+	hbKey := serviceKey(sid)
```
cw9: Does serviceKey() include zone? perhaps we need to differentiate zones as well.

PR #90, opened 2017-02-24
Fix bug in backing out of a replace

cw9:
Previously when backing out from a replace, the empty instance in the
placement is not being cleaned up, this diff fixes that.

@xichen2020 @robskillington

### DIFF
--- a/services/placement/algo/sharded_helper.go
+++ b/services/placement/algo/sharded_helper.go
@@ -30,13 +30,17 @@ import (
 	"github.com/m3db/m3cluster/shard"
 )
 
+type includeInstanceType int
+
+const (
+	includeEmpty includeInstanceType = iota
+	nonEmptyOnly
+)
+
 // PlacementHelper helps the algorithm to place shards
 type PlacementHelper interface {
 	// PlaceShards distributes shards to the instances in the helper, with aware of where are the shards coming from
 	PlaceShards(shards []shard.Shard, from services.PlacementInstance) error
-
-	// GeneratePlacement generates a placement
-	GeneratePlacement(includeEmptyInstance bool) services.ServicePlacement
 
 	// TargetLoadForInstance returns the targe load for a instance
 	TargetLoadForInstance(id string) int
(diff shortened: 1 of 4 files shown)


services/placement/algo/sharded_helper.go
```diff
@@ -134,15 +134,21 @@ func (ph *placementHelper) BuildInstanceHeap(instances []services.PlacementInsta
 	return newHeap(instances, availableCapacityAscending, ph.targetLoad, ph.rackToWeightMap)
 }
 
-func (ph *placementHelper) GeneratePlacement() services.ServicePlacement {
-	instancesWithLoad := make([]services.PlacementInstance, 0, len(ph.instances))
-	for _, instance := range ph.instances {
-		if loadOnInstance(instance) > 0 {
-			instancesWithLoad = append(instancesWithLoad, instance)
+func (ph *placementHelper) GeneratePlacement(includeEmptyInstance bool) services.ServicePlacement {
+	var instances []services.PlacementInstance
```
xichen2020: nit: it's slightly simpler to do
```
instances := ph.instanceList()
if !includeEmptyInstance {
   instances = make([]services.PlacementInstance, 0, len(ph.instances))
   ...
}
```

services/placement/algo/sharded_helper.go
```diff
@@ -36,7 +36,7 @@ type PlacementHelper interface {
 	PlaceShards(shards []shard.Shard, from services.PlacementInstance) error
 
 	// GeneratePlacement generates a placement
-	GeneratePlacement() services.ServicePlacement
+	GeneratePlacement(includeEmptyInstance bool) services.ServicePlacement
```
xichen2020: Perhaps it's better to make it an enum so it's easier to understand at callsite:
```
type includeInstanceType int

const (
  includeEmpty includeInstanceType = iota
  nonEmptyOnly
)

GeneratePlacement(instanceType includeInstanceType) service.ServicePlacement
```
Then it's easier to understand what it means at callsite
```
GeneratePlacement(includeEmpty)
```

cw9: sounds good

services/placement/algo/sharded_helper.go
cw9: feels not so idiomatic after switching to the switch/enum way

PR #91, opened 2017-02-25
Call Get() after service discovery watch timed-out for initial value

cw9:
So that the config for service discovery watch timeout will always try a Get() and is independent
from the config for watch creation timeout and/or get request timeouts

@xichen2020 @robskillington

### DIFF
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -47,13 +47,13 @@ var (
 )
 
 // NewStore creates a kv store based on etcd
-func NewStore(c *clientv3.Client, opts Options) (kv.Store, error) {
+func NewStore(etcdKV clientv3.KV, etcdWatcher clientv3.Watcher, opts Options) (kv.Store, error) {
 	scope := opts.InstrumentsOptions().MetricsScope()
 
 	store := &client{
 		opts:           opts,
-		kv:             c.KV,
-		watcher:        c.Watcher,
+		kv:             etcdKV,
+		watcher:        etcdWatcher,
 		watchables:     map[string]kv.ValueWatchable{},
 		retrier:        xretry.NewRetrier(opts.RetryOptions()),
 		logger:         opts.InstrumentsOptions().Logger(),
@@ -69,7 +69,7 @@ func NewStore(c *clientv3.Client, opts Options) (kv.Store, error) {
 		},
 	}
 	wOpts := watchmanager.NewOptions().
-		SetWatcher(c.Watcher).
+		SetWatcher(etcdWatcher).
 		SetUpdateFn(store.update).
 		SetTickAndStopFn(store.tickAndStop).
 		SetWatchOptions([]clientv3.OpOption{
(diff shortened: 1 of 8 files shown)


kv/etcd/store.go
```diff
@@ -522,8 +522,13 @@ func newValue(val []byte, ver, rev int64) *value {
 	}
 }
 
-func (c *value) isNewer(other *value) bool {
-	return c.Rev > other.Rev
+func (c *value) IsNewer(other kv.Value) bool {
+	othervalue, ok := other.(*value)
```
xichen2020: Is there a case where `kv.Value` is not a `value` pointer for the etcd store?

services/client/services.go
```diff
@@ -397,7 +412,15 @@ func (c *client) watchPlacementAndHeartbeat(
 			if value == nil {
 				// NB(cw) this can only happen when the placement has been deleted
 				// it is safer to let the user keep using the old topology
-				c.logger.Infof("received placement update with nil value")
```
xichen2020: This looks the same as the logic above in `watchPlacement`. Can we move them into a method so for future changes we don't need to make the same change in two places?

services/client/services.go
```diff
@@ -249,12 +249,17 @@ func (c *client) Watch(sid services.ServiceID, opts services.QueryOptions) (xwat
...
-	placementWatch, err := initPlacementWatch(kvm.kv, sid, c.opts.InitTimeout())
+	placementWatch, err := kvm.kv.Watch(placementKey(sid))
 	if err != nil {
 		return nil, err
 	}
 
-	initService, err := getServiceFromValue(placementWatch.Get(), sid)
+	initValue, err := waitForInitValue(kvm.kv, placementWatch, sid, c.opts.InitTimeout())
+	if err != nil {
+		return nil, fmt.Errorf("could not get init value within timeout, err: %s", err)
```
xichen2020: Should be `%v` here?

kv/etcd/store.go
cw9: nope, just wanna make the check safer

services/client/services.go
cw9: yeah good catch

kv/etcd/store.go
xichen2020: If so, I'd make it a type assertion so it'll panic if the value doesn't match what's expected instead of swallowing it silently

services/client/services.go
cw9: yeah sounds good

kv/etcd/store.go
cw9: I'd rather keep it this way in case the user uses 2 different kv stores, say etcd + cassandra or etcd + flipr, and different impls of kv.Value may exist in the runtime, it's probably safer not to just panic

xichen2020: So does that mean `kv.Value` can be some type other than a `*value`? We should only keep the `ok` check if and only if the `kv.Value` can be more than one type.

cw9: yeah for example if a user decides to compare this *value and *mem.value (although no one really should do that) this check would come in handy.

xichen2020: Ok as long as there is a legitimate case where the value can be more than one type, I'm fine with it

PR #95, opened 2017-03-03
[advertisement] Tick immediately on start

schallert:
When we start advertising it takes as long as the heartbeat interval
duration to begin advertising. Ideally a service shouldn't start
advertising until it's ready, so when it starts it might make sense to
show the advertisement immediately. Even if it's not ready at first, if
it has a health function it won't start heartbeating until it's ready
anyway.

RFR @cw9

### DIFF
--- a/services/client/services.go
+++ b/services/client/services.go
@@ -153,16 +153,22 @@ func (c *client) Advertise(ad services.Advertisement) error {
 		sid := ad.ServiceID()
 		errCounter := c.serviceTaggedScope(sid).Counter("heartbeat.error")
 
+		tickFn := func() {
+			if isHealthy(ad) {
+				if err := hb.Heartbeat(pi, m.LivenessInterval()); err != nil {
+					c.logger.Errorf("could not heartbeat service %s, %v", sid.String(), err)
+					errCounter.Inc(1)
+				}
+			}
+		}
+
+		tickFn()
+
 		ticker := time.Tick(m.HeartbeatInterval())
 		for {
 			select {
 			case <-ticker:
-				if isHealthy(ad) {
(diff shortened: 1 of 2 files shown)


services/client/services.go
```diff
@@ -153,16 +153,26 @@ func (c *client) Advertise(ad services.Advertisement) error {
...
+
+		tickFn := func() {
+			if isHealthy(ad) {
+				if err := hb.Heartbeat(pi, m.LivenessInterval()); err != nil {
+					c.logger.Errorf("could not heartbeat service %s, %v", sid.String(), err)
+					errCounter.Inc(1)
+				}
+			}
+		}
+
```
cw9: just call tickFn() before the for loop?

services/client/services_test.go
```diff
@@ -136,6 +136,16 @@ func TestAdvertiseErrors(t *testing.T) {
 	err = sd.Advertise(ad)
 	require.NoError(t, err)
 
+	hbGen.Lock()
+	hb := hbGen.hbs[serviceKey(sid)]
+	hbGen.Unlock()
+
+	// hb store should show advertisement (almost) immediately
+	time.Sleep(time.Millisecond)
```
cw9: can you try run this test with --count 1000? feeling this is gonna be flaky

cw9: the test case could be: construct the sd client with a super long HeartbeatInterval, like an hour, and verify the heartbeat still got created pretty soon

services/client/services.go
schallert: Updated!

services/client/services_test.go
schallert: Increased the relevant times to give the tests a bit more headroom. Ran with `-c 100` and didn't see any flakes so far.

PR #97, opened 2017-03-27
Add cutover time to placement proto

xichen2020:
cc @cw9 @robskillington 

This PR adds cutover time to the placement proto message so clients receiving placement updates can coordinate when the new placement should take effect if needed.


cw9:
The logic is coming in a different pr?

xichen2020:
Yes, the coordination logic will be on the client side and doesn't live inside m3cluster

PR #98, opened 2017-04-03
Cancel context in etcd.Watch() call when there is a timeout

cw9:
@xichen2020


etcd/watchmanager/manager.go
```diff
@@ -69,10 +69,12 @@ type metrics struct {
 func (w *manager) watchChanWithTimeout(key string) (clientv3.WatchChan, error) {
 	doneCh := make(chan struct{})
 
+	ctx, cancelCtx := context.WithCancel(clientv3.WithRequireLeader(context.Background()))
```
xichen2020: nit: `cancelFn` perhaps?

cw9: sure

PR #99, opened 2017-04-10
Add placement snapshots for staged cutover

xichen2020:
cc @cw9 @jeromefroe @martin-mao @robskillington


generated/proto/placement/placement.proto
```diff
@@ -49,3 +49,7 @@ enum shard_state {
   Available = 1;
   Leaving = 2;
 }
+
+message placementSnapshots {
```
cw9: should the convention be placement_snapshots?

xichen2020: The style guide says it should be `PlacementSnapshots`, but the existing proto message names didn't follow that convention, so it's either that or being inconsistent with the existing names in the repo. I chose to be consistent in this case.

PR #100, opened 2017-04-10
Add interface for Transaction Store

cw9:


kv/types.go
```diff
@@ -105,3 +105,66 @@ type Store interface {
...
+	// SetType sets the type of the operation
+	SetType(ot OpType) Op
+
+	// Key returns the key used in the operation
+	Key() string
+	// SetKey sets the key in the operation
+	SetKey(key string) Op
+
+	// Value returns the value to be set, if the op type is Set
+	Value() proto.Message
```
xichen2020: Talked offline, let's move this outside the `Op` interface and into perhaps a type-specific interface (e.g., `ValueOp` and `NoValueOp`).

kv/store.go
```diff
@@ -20,7 +20,10 @@
 
 package kv
 
-import xwatch "github.com/m3db/m3x/watch"
+import (
+	"github.com/gogo/protobuf/proto"
```
xichen2020: This is not the correct import, you want `"github.com/golang/protobuf/proto"`

kv/store.go
```diff
@@ -88,3 +91,44 @@ func valueFromWatch(value interface{}) Value {
...
+}
+
+type opResponse struct {
+	opBase
+
+	value interface{}
+}
+
+// NewOpResponse returns a OpResponse
+func NewOpResponse(key string, value interface{}) OpResponse {
```
xichen2020: Does this ctor need to be exported?

kv/store.go
```diff
@@ -20,7 +20,10 @@
 
 package kv
```
xichen2020: Might wanna pull op related stuff into `op.go`

cw9: will do

kv/store.go
cw9: yeah probably private is enough

PR #101, opened 2017-04-11
Add implementation of TxnStore

cw9:
@xichen2020 @robskillington

### DIFF
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -47,7 +47,7 @@ var (
 )
 
 // NewStore creates a kv store based on etcd
-func NewStore(etcdKV clientv3.KV, etcdWatcher clientv3.Watcher, opts Options) (kv.Store, error) {
+func NewStore(etcdKV clientv3.KV, etcdWatcher clientv3.Watcher, opts Options) (kv.TxnStore, error) {
 	scope := opts.InstrumentsOptions().MetricsScope()
 
 	store := &client{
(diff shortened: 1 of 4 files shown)


kv/etcd/store.go
```diff
@@ -234,6 +234,82 @@ func (c *client) History(key string, from, to int) ([]kv.Value, error) {
...
+			etcdOps[i] = clientv3.OpPut(
+				c.opts.ApplyPrefix(opSet.Key()),
+				string(value),
+				clientv3.WithPrevKV(),
+			)
+
+			// NB(cw) We need to initiate the OpResponse with a default version here
+			// because if a key was set first time, there is no way to get the key
+			// in the response from etcd
+			opResponses[i] = kv.NewOpResponse(opSet.Key(), 1)
```
xichen2020: Might be better to set the version when processing the response?

kv/etcd/store.go
```diff
@@ -234,6 +234,82 @@ func (c *client) History(key string, from, to int) ([]kv.Value, error) {
...
+		return nil, err
+	}
+	if !r.Succeeded {
+		return nil, kv.ErrConditionCheckFailed
+	}
+
+	for i := range r.Responses {
+		// NB(cw) once we start to support other OpType
+		// we need to check the response type
+		res := r.Responses[i].GetResponsePut()
```
xichen2020: Can we check the response type here and return an error if it's not a put response for forward compatibility?

cw9: unfortunately there is no good way to check the type, can only check if the GetResponsePut() returns nil or not

kv/etcd/store.go
cw9: sounds good

kv/etcd/store.go
```diff
@@ -242,48 +284,30 @@ func (c *client) Txn(conditions []kv.Condition, ops []kv.Op) (kv.Response, error
...
-			)
-
-			// NB(cw) We need to initiate the OpResponse with a default version here
-			// because if a key was set first time, there is no way to get the key
-			// in the response from etcd
-			opResponses[i] = kv.NewOpResponse(opSet.Key(), 1)
-		default:
-			return nil, kv.ErrUnknownOpType
-		}
+		// NB(cw) We need to initiate the OpResponse with a default version here
```
xichen2020: Move or remove this comment?

PR #102, opened 2017-04-11
Expose TxnStore from m3cluster client

cw9:
@xichen2020

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -33,4 +33,7 @@ type Client interface {
 
 	// KV returns access to the distributed configuration store
 	KV() (kv.Store, error)
+
+	// Txn returns access to the transaction store
+	Txn() (kv.TxnStore, error)
 }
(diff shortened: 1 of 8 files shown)


client/etcd/client.go
```diff
@@ -98,7 +98,15 @@ func (c *csclient) Services() (services.Services, error) {
 
 func (c *csclient) KV() (kv.Store, error) {
 	c.kvOnce.Do(func() {
```
xichen2020: Might be better to move the logic inside `KV` and `Txn` into an unexported internal function?

client/client.go
```diff
@@ -33,4 +33,7 @@ type Client interface {
 
 	// KV returns access to the distributed configuration store
 	KV() (kv.Store, error)
+
+	// Txn returns access to the transaction store
+	Txn() (kv.TxnStore, error)
```
xichen2020: Would it better to call it `TxnStore`? `Txn` feels like it should return a `transaction` object

client/etcd/client.go
cw9: sounds good

client/client.go
cw9: sounds good

PR #104, opened 2017-04-11
Migrate kvutil from statsdex

cw9:


generated/proto/commonpb/common.proto
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
xichen2020: Perhaps just call it common/?

kv/util/util.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/m3db/m3x/log"
+)
+
+var errNilStore = errors.New("kv store is nil")
+
+type getValueFn func(kv.Value) (interface{}, error)
+type updateFn func(interface{})
+
+// WatchAndUpdateBool sets up a watch for a bool property.
+func WatchAndUpdateBool(store kv.Store, name string, property *bool, lock sync.Locker,
```
xichen2020: I wonder if this is a good candidate to explore a template based approach to generate code blocks that only differ by their types.

cw9: sure thing

generated/proto/commonpb/common.proto
cw9: I actually prefer calling it commonpb since there is a higher chance of collision with another common package, one thing I hate about the current package name in statsdex for those protos is that I always have to manually rename the import.

xichen2020: Sure thing, in that case you can also name the directory "common" and have the package name be "commonpb", i don't feel strongly about it though

kv/util/util.go
```diff
@@ -16,144 +16,101 @@ type getValueFn func(kv.Value) (interface{}, error)
 type updateFn func(interface{})
 
 // WatchAndUpdateBool sets up a watch for a bool property.
-func WatchAndUpdateBool(store kv.Store, name string, property *bool, lock sync.Locker,
+func WatchAndUpdateBool(store kv.Store, key string, property *bool, lock sync.Locker,
 	defaultValue bool, logger xlog.Logger) error {
-	if store == nil {
-		return errNilStore
-	}
+	update := updateWithLock(func(i interface{}) { *property = i.(bool) }, lock)
```
xichen2020: The convention is to suffix function vars with `Fn`.

generated/proto/commonpb/common.proto
cw9: Talked offline gonna keep the package name and the directory name in sync to make IDE auto complete easier

kv/util/util.go
```diff
@@ -0,0 +1,213 @@
...
+
+	updateWithKV(getInt64, updateFn, key, v, defaultValue, logger)
+
+	return res
+}
+
+// StringArrayFromValue gets a string array from kv.Value
+func StringArrayFromValue(v kv.Value, key string, defaultValue []string, logger xlog.Logger) []string {
+	var res []string
+	updateFn := func(i interface{}) { res = i.([]string) }
```
prateek: Does it make sense to change the updateFn to return error for invalid type conversions, i.e.

```
type updateFn func(i interface{}) error

func StringArrayFromValue(...) []string {
  var res []string
  updateFn := func(i interface{}) error {
      res, ok := i.([]string)
      if !ok { return errValueIsNotAStringArray }
      return nil
  }
}
```

cw9: it will never happen, the interface of getValueFn and updateFn has to use interface{} since the lack of generic in Go, but within each usage, the type in getValueFn and updateFn should be the same, so the cast is safe

prateek: makes sense, LGTM

PR #106, opened 2017-04-14
Add runtime value interface to m3cluster

xichen2020:
cc @cw9 @jeromefroe @martin-mao 

This PR adds the runtime value interface to m3cluster, which provides the ability to watch / unwatch updates and process the updates using the configurable unmarshalling function and processing function.

### DIFF
--- a/kv/util/runtime/options.go
+++ b/kv/util/runtime/options.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


kv/util/runtime/value_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	_, rv := testValueWithMockStore(ctrl)
+	rv.version = 3
+	rv.unmarshalFn = func(v kv.Value) (interface{}, error) { return v, nil }
+
+	require.NoError(t, rv.updateWithLock(mockValue{version: 2}))
+	require.Equal(t, 3, rv.version)
+}
+
+type mockValue struct {
```
cw9: can this be replaced with mem.NewValue() ?

xichen2020: Sure thing

xichen2020: Sure thing.

kv/util/runtime/value.go
```diff
@@ -0,0 +1,208 @@
...
+	"sync"
+	"time"
+
+	"github.com/m3db/m3cluster/kv"
+	"github.com/m3db/m3x/log"
+)
+
+const (
+	// DefaultVersion is the default version of an uninitialized kv value.
+	DefaultVersion = 0
```
xichen2020: Talked offline, will move to `kv/`

kv/util/runtime/options.go
```diff
@@ -0,0 +1,133 @@
...
+	unmarshalFn      UnmarshalFn
+	processFn        ProcessFn
+}
+
+// NewOptions creates a new set of options.
+func NewOptions() Options {
+	return &options{
+		instrumentOpts:   instrument.NewOptions(),
+		initWatchTimeout: defaultInitWatchTimeout,
+		kvStore:          mem.NewStore(),
```
xichen2020: Hm using the memstore feels like a reasonable default to me, leaving it nil is slightly better but we are back to the same place if the users don't call `Validate`.

PR #107, opened 2017-04-16
Migrate config for etcd based config service client

cw9:

### DIFF
--- a/client/etcd/config.go
+++ b/client/etcd/config.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package etcd
+
+import "time"
+import "github.com/m3db/m3cluster/client"
+
+// M3SDConfiguration is the config for service discovery
+type M3SDConfiguration struct {
+	InitTimeout time.Duration `yaml:"initTimeout"`
+}
+
+// ClusterConfig is the config for a zoned etcd cluster
+type ClusterConfig struct {
+	Zone      string   `yaml:"zone"`
+	Endpoints []string `yaml:"endpoints"`
+}
+
+// Configuration is for config service client
+type Configuration struct {
+	Zone         string            `yaml:"zone"`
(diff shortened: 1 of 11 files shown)


client/etcd/config.go
```diff
@@ -0,0 +1,71 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package etcd
+
+import "time"
```
xichen2020: Use an import block perhaps?

client/etcd/config.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// NewClient creates a new config service client
+func (cfg Configuration) NewClient() (client.Client, error) {
+	return NewConfigServiceClient(cfg.options())
+}
+
+func (cfg Configuration) options() Options {
+	return NewOptions().
+		SetZone(cfg.Zone).
```
xichen2020: What happens if any of these properties isn't set? Should we let those default to the default values in the options ? Or perhaps make them nonzero if not?

glide.lock
```diff
@@ -129,7 +130,7 @@ imports:
   subpackages:
   - prometheus
 - name: github.com/prometheus/client_model
-  version: fa8ad6fec33561be4280a8f0514318c79d7f6cb6
+  version: 6f3806018612930941127f2a7c6c453ba2c527d2
```
xichen2020: Can we perhaps pin these versions?

client/etcd/config.go
cw9: sure

client/etcd/config.go
cw9: it's fine if any of them are empty, NewClient will always check options.Validate()

glide.lock
cw9: sure

cw9: actually since this is only used in etcd package, we should probably not pin it, so that etcd can pin it to a version if needed. If we pin it now, we might forgot why we pinned it in the first place and hard to unpin in the future.

client/etcd/config.go
xichen2020: Seems `Validate` doesn't check if `env` or `zone` is set, is it okay for them to be empty?

glide.lock
xichen2020: We usually prefer deterministic behavior by pinning the package versions so it is guaranteed that glide update will always install the same versions. As a side benefit, it also reduces noises when someone run glide update next time.

xichen2020:
Seems there's a deadlock situation `TestWatchRecreateWatch`?

client/etcd/config.go
cw9: Yeah they are OK to be empty

client/etcd/options.go
```diff
@@ -35,14 +35,14 @@ type Options interface {
 	Zone() string
 	SetZone(z string) Options
 
-	ServiceInitTimeout() time.Duration
-	SetServiceInitTimeout(timeout time.Duration) Options
+	Service() string
```
prateek: Is `ServiceID` clearer than `Service`?

cw9: ServiceID usually means [this](https://github.com/m3db/m3cluster/blob/master/services/types.go#L134) in the kv world, so I'd rather not to confuse with it.

prateek: Ah. Looks like this interface exposes `Zone`, `Env` and `Name`; instead of that, could it have a ServiceID field?

cw9: Yeah I thought about that, I didn't do that because user of this Option does not necessarily need to use the service discovery client which contains the idea of ServiceID, they can just use the KV interface, also the validation for this Option only requires the Service field to be non-empty, while ServiceID has different validation rules so I can't call serviceID.Validate() here but have to call `opts.ServiceID().Service() != ""` which is a bit weird

prateek: Yea, just feels weird having a type for precisely these fields and not using it. Why is it valid for these options to not have an env/zone - does that query multiple zones/envs for properties?

prateek: spoke to Chao offline, Service here is used as an identifier for the filename generation and metrics; doesn't mean the same thing as Name from ServiceID.

PR #108, opened 2017-04-17
Change placement proto convention

cw9:
Tested to be compatible with old proto

@xichen2020 @prateek @robskillington


generated/proto/placement/placement.proto
```diff
@@ -21,35 +21,35 @@ syntax = "proto3";
...
+  ShardState state = 2;
   string source_id = 3;
 }
 
-enum shard_state {
-  Initializing = 0;
-  Available = 1;
-  Leaving = 2;
+enum ShardState {
+  INITIALIZING = 0;
```
prateek: It's recommended to have 'UNKNOWN' as the first value in an enum. This will cause older clients to default to 'UNKNOWN' if they encounter new enum values. If we don't have active users of this type, could you make that change too.

cw9: sounds good, will change

prateek: Was afraid of that :( We'll live with it for now, opened an issue against this repo so can address in the future - https://github.com/m3db/m3cluster/issues/110

xichen2020:
LGTM, I guess the type names used in m3collector and m3db will need to change accordingly, though.

PR #109, opened 2017-04-18
Refactor config service configuration struct

cw9:
When timeout is not positive, a default timeout will be applied, so
unnecessary to validate it here.

@xichen2020 @prateek

### DIFF
--- a/services/client/config.go
+++ b/services/client/config.go
@@ -4,14 +4,14 @@ import "time"
 
 // Configuration is the config for service discovery
 type Configuration struct {
-	InitTimeout *time.Duration `yaml:"initTimeout"`
+	InitTimeout time.Duration `yaml:"initTimeout"`
 }
 
 // NewOptions creates an Option
 func (cfg Configuration) NewOptions() Options {
 	opts := NewOptions()
-	if cfg.InitTimeout != nil {
-		opts = opts.SetInitTimeout(*cfg.InitTimeout)
+	if cfg.InitTimeout != 0 {
+		opts = opts.SetInitTimeout(cfg.InitTimeout)
 	}
 	return opts
 }
--- a/services/client/config.go
+++ b/services/client/config.go
@@ -1,0 +1,17 @@
+package client
+
+import "time"
+
+// Configuration is the config for service discovery
+type Configuration struct {
+	InitTimeout *time.Duration `yaml:"initTimeout"`
+}
+
+// NewOptions creates an Option
+func (cfg Configuration) NewOptions() Options {
+	opts := NewOptions()
+	if cfg.InitTimeout != nil {
+		opts = opts.SetInitTimeout(*cfg.InitTimeout)
+	}
+	return opts
+}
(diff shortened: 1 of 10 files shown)


prateek:
I'm not sure I follow - the following test code causes the old Validate() to fail, but the new one to succeed. Is that the intent?

```
negativeMinute := -1 * time.Minute
err := NewOptions().SetInitTimeout(negativeMinute).Validate()
require.NoError(t, err)
```

cw9:
yes that's the intent, in the code where the value of opts.InitTimeout() is used, it will use to valid value when -1 * time.Minute is returned, I could change the InitTimeout() to return the default value when non positive value is set as an alternative.

prateek:
If -1*time.Minute should not be used, I'd expect Validate() to complain about it. Otherwise what's the point of calling Validate()?

cw9:
The reason I'm removing it is to save the user from providing a value in the config yaml for this field if he or she does not really want to override it. If we want to enforce the users to provide a valid value I'm fine with dropping this diff.

With the validation the empty config will fail since it tries to set it to 0, the Validate() checks the store to be nil so we won't panic on using the stores, but the -1 minute can be and has already been handled by the code by using a valid value instead.

xichen2020:
It might be better to check the timeout value in the config and don't set it if the value is not positive when parsing the configs, as opposed to removing the validation check.

cw9:
It also feels kinda weird cause it's like pushing the logic in sd implementation up here. But I don't feel that strongly either way

prateek:
How about nil checking during config parsing? 

i.e. change config.NewOptions to 
```
func (cfg Configuration) NewOptions() Options {
     opts := NewOptions().
		SetZone(cfg.Zone).
		SetEnv(cfg.Env).
		SetService(cfg.Service).
		SetCacheDir(cfg.CacheDir).
		SetClusters(cfg.etcdClusters())
    if timeout := cfg.M3SD.InitTimeout; timeout != nil {
       opts = opts.SetServiceInitTimeout(timeout) 
    }
    return opts
}
```

Would that allow CreateServices to remain un-changed?

cw9:
> How about nil checking during config parsing?

Yeah that sounds good, refactored the config stuff a bit

services/client/config.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Configuration is the config for service discovery
+type Configuration struct {
+	InitTimeout *time.Duration `yaml:"initTimeout"`
+}
+
+// NewOptions creates an Option
+func (cfg Configuration) NewOptions() Options {
+	opts := NewOptions()
+	if cfg.InitTimeout != nil {
```
xichen2020: Is it okay if the config explicitly set the timeout to 0?

cw9: yeah, it will just fail options.Validation()

xichen2020: In that case why not just make `InitTimeout` a `time.Duration` and don't call `SetInitTimeout` if `cfg.InitTimeout == 0`?

cw9: yeah sounds good, removed the pointer

PR #111, opened 2017-04-19
Update Placement Service Interface

prateek:
- update travis configs to meet expectations from ci-scripts
- expose instance used information from PlacementService 

/cc  @cw9 @xichen2020

### DIFF
--- a/services/placement/service/placementservice.go
+++ b/services/placement/service/placementservice.go
@@ -147,30 +147,30 @@ func (ps placementService) AddReplica() (services.ServicePlacement, error) {
 
 func (ps placementService) AddInstance(
 	candidates []services.PlacementInstance,
-) (services.ServicePlacement, error) {
+) (services.ServicePlacement, services.PlacementInstance, error) {
 	p, v, err := ps.ss.Placement(ps.service)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	var addingInstance services.PlacementInstance
 	if addingInstance, err = ps.findAddingInstance(p, candidates, ps.opts); err != nil {
-		return nil, err
+		return nil, nil, err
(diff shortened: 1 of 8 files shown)


services/placement/service/placementservice.go
```diff
@@ -147,30 +147,30 @@ func (ps placementService) AddReplica() (services.ServicePlacement, error) {
...
 
 	if err := placement.Validate(p); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	if ps.opts.Dryrun() {
 		ps.logger.Info("this is a dryrun, the operation is not persisted")
-		return p, err
+		return p, addingInstance, err
```
cw9: this would return you the addingInstance with empty shards, you should probably return p.Instance(addingInstance.ID()) to get the added instance with shard assignment

prateek: feels weird to return the identifier instead of the object. Is there a way to add the shard assignment to the instance?

cw9: Sorry I meant to return the instance object by looking up the addingInstance.ID() in the new placement

prateek: ah cheers, will address

prateek: done, take a look

services/placement/service/placementservice.go
```diff
@@ -165,6 +165,11 @@ func (ps placementService) AddInstance(
 		return nil, nil, err
 	}
 
+	addingInstance, ok := p.Instance(addingInstance.ID())
+	if !ok {
+		return nil, nil, fmt.Errorf("unable to find added instance in new placement")
```
cw9: add the instance id in the error msg?

PR #113, opened 2017-04-26
[client/etcd] Refactor clientinterface to require a namespace

jeromefroe:
@cw9 @xichen2020 @schallert This PR refactors the `Client` interface to require a namespace when calling `KV` and `Txn`. This is so we can easily support different namespaces moving forward, e.g. `/kv` or `/r2`, that reside in the same etcd cluster per our discussions earlier.

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -42,7 +42,6 @@ import (
 const (
 	keySeparator    = "/"
 	cacheFileFormat = "%s-%s.json"
-	kvPrefix        = "_kv/"
 )
 
 type newClientFn func(endpoints []string) (*clientv3.Client, error)
@@ -94,14 +93,14 @@ func (c *csclient) Services() (services.Services, error) {
 	return c.sd, c.sdErr
 }
 
-func (c *csclient) KV() (kv.Store, error) {
-	c.createTxnStore()
+func (c *csclient) KV(namespace string) (kv.Store, error) {
+	c.createTxnStore(namespace)
 
 	return c.kv, c.kvErr
 }
 
-func (c *csclient) Txn() (kv.TxnStore, error) {
-	c.createTxnStore()
+func (c *csclient) Txn(namespace string) (kv.TxnStore, error) {
+	c.createTxnStore(namespace)
 
 	return c.kv, c.kvErr
 }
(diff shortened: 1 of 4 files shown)


client/etcd/client.go
```diff
@@ -212,12 +211,13 @@ func fileName(service, zone string) string {
 	return strings.Replace(cacheFileName, string(os.PathSeparator), "_", -1)
 }
 
-func prefix(env string) string {
-	res := kvPrefix
+func fullPrefix(namespace string, env string) string {
+	prefix := concat(keySeparator, concat(namespace, keySeparator))
```
xichen2020: nit: it might be readable to just write it as
```
p := namespace
if env != "" {
   p += keySeparator + env
}
return keySeparator + p + keySeparator
```

glide.lock
```diff
@@ -134,7 +134,7 @@ imports:
   subpackages:
   - go
 - name: github.com/prometheus/common
-  version: 195bde7883f7c39ea62b0d92ab7359b5327065cb
+  version: 9e0844febd9e2856f839c9cb974fbd676d1755a8
```
xichen2020: Why do we need to update prometheus?

client/etcd/client.go
```diff
@@ -42,7 +42,6 @@ import (
 const (
 	keySeparator    = "/"
 	cacheFileFormat = "%s-%s.json"
-	kvPrefix        = "_kv/"
```
cw9: Please update statsdex to make sure they are still picking up the current keys

xichen2020:
Also something worth checking out: https://godoc.org/github.com/coreos/etcd/clientv3/namespace. The package is a clientv3 wrapper that translates all keys to begin with a given prefix.

glide.lock
jeromefroe: It's not in `glide.yaml` so it was bumped when I ran an update. I'll put it in `glide.yaml` to pin it.

jeromefroe:
I like that approach. Perhaps we don't set the prefix but instead provide wrapper functions to add the namespace to a `Store` and `TxnStore`?

jeromefroe:
This refactoring is, in fact, more involved than I first realized since service discovery and heartbeat already use this approach. Therefore, I'm going to close this PR and open up an issue so we can return to this at a later, more convenient, time.

xichen2020:
I'm fine w/ punting it for now, but does it mean if we use kv.Store to set values, the key will still be under _kv/?

jeromefroe:
Just for auditing purposes: we talked offline and agreed to use the `_kv/` namespace for now.

PR #115, opened 2017-05-03
Add LeaderService for service elections

schallert:
Adds `LeaderService` to do elections across service instances. See `TESTING.md` for description of integration testing methodology.

**NOTE**: since it doesn't appear until the end of the diff, reading all the comments in `services/types.go` before reading the code will aid in context some


services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+	errNilElection = errors.New("nil election was returned")
+)
+
+// NewService creates a new leader service client based on an etcd client.
+func NewService(cli *clientv3.Client, opts Options) (services.LeaderService, error) {
+	if err := opts.Validate(); err != nil {
+		return nil, err
+	}
+
+	sessionOpts := []concurrency.SessionOption{}
```
cw9: nit: you can do 
`var sessionOpts []concurrency.SessionOption`

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	election := concurrency.NewElection(session, servicePrefix(opts.ServiceID()))
+	if election == nil {
+		return nil, errNilElection
+	}
+
+	c := &client{
```
cw9: nit: return &client{...}, nil

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+func (c *client) Leader() (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), leaderCallTimeout)
+	defer cancel()
+	return c.election.Leader(ctx)
+}
+
+// Close closes the election service client entirely. No more campaigns can be
+// started and any outstanding campaigns are closed.
+func (c *client) Close() error {
+	return c.session.Close()
```
cw9: do you need to clean up c.ctx and c.ctxCancel?

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+	election  *concurrency.Election
+	session   *concurrency.Session
+	val       string
+}
+
+func (c *client) Campaign() error {
+	ctx, cancel := context.WithCancel(context.Background())
```
cw9: check if c is closed?

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (c *client) Resign() error {
+	c.Lock()
+	defer c.Unlock()
+
+	// if we're not the leader but still campaigning, cancelling the context
+	// will stop the campaign
+	if c.ctx != nil && c.ctxCancel != nil {
+		c.ctxCancel()
+		c.ctx = nil
```
cw9: do you need to keep the ctx in the client?

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// electionValue returns the OverrideValue from opts if it is non-empty,
+// otherwise returns the current hostname by default.
+func electionValue(opts Options) string {
+	if o := opts.OverrideValue(); o != "" {
+		return o
+	}
+
+	h, err := os.Hostname()
```
cw9: reuse this by setting it to be the default OverrideValue?

services/leader/service.go
```diff
@@ -0,0 +1,144 @@
...
+	if ttl := opts.TTL(); ttl != 0 {
+		sessionOpts = append(sessionOpts, concurrency.WithTTL(ttl))
+	}
+
+	session, err := concurrency.NewSession(cli, sessionOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	election := concurrency.NewElection(session, servicePrefix(opts.ServiceID()))
```
cw9: when would they return nil election?

services/types.go
```diff
@@ -326,3 +330,22 @@ type HeartbeatService interface {
...
+
+// LeaderService provides access to etcd-backed leader elections.
+type LeaderService interface {
+	// Close closes the election service client entirely. No more campaigns can be
+	// started and any outstanding campaigns are closed.
+	Close() error
+
+	// Campaign proposes that the caller become the leader, blocking until it is
+	// elected.
+	Campaign() error
```
cw9: We need to support campaigning for multiple locks, for example: if we want to support master/slave mode per shard, then a box might be owning 20 shards, and became the leader for 15 shards and slave for 5 shards. We should allow services to content for any keys under the service prefix, e.g: there could be _ld/env/svc/1, _ld/env/svc/2 (for shards), _ld/env/svc/bucket1 (bucket1 here means a subset of shards, as long as the service understands what it means), a service can create any lock under the service prefix.

Also I wonder if we should name the functions as lock/unlock to fit better into that semantic.

integration/leader_test.go
```diff
@@ -0,0 +1,145 @@
...
+		SetZone("z1").
+		SetName("s1")
+
+	return leader.NewOptions().
+		SetServiceID(sid).
+		SetOverrideValue(override).
+		SetTTL(ttl)
+}
+
+func TestIntegration_Simple(t *testing.T) {
```
cw9: Wonder what happens if you create a client on the test cluster and shutdown the cluster, after long enough time (past ttl) can the client still keep the lock?

services/leader/service.go
schallert: Turns out never! Deleted

schallert:
@cw9 I modified the APIs to allow to specify a sub-election when creating the `LeaderService` instance so that we can have multiple elections per service, while still supporting service-wide elections. I kept the scope at creation time for the service, since I didn't want to have to pass the election ID to every call, since that sounds messy (and we need separate etcd sessions per-election anyway). I'm open to changing the function names if we want, though if the scope of the sub-election is at the service level I think the current names still fit

services/leader/client.go
```diff
@@ -0,0 +1,229 @@
...
+// started and any outstanding campaigns are closed.
+func (c *client) Close() error {
+	if c.close() {
+		c.Lock()
+		c.cancelWithLock()
+		c.Unlock()
+
+		c.wb.Update(CampaignClosed)
+		c.wb.Close()
+		return c.session.Close()
```
cw9: when does session.Close() error out? is it ok that the client is marked as closed but session errored out ?

services/leader/client_test.go
```diff
@@ -0,0 +1,269 @@
...
+	_, err = svc.Campaign()
+	assert.Equal(t, ErrCampaignInProgress, err)
+}
+
+func TestResign(t *testing.T) {
+	tc := newTestCluster(t)
+	defer tc.close()
+
+	svc := tc.client("i1")
+
```
cw9: What happens if you call Resign() before Campaign()?

services/leader/client.go
```diff
@@ -0,0 +1,229 @@
...
+		c.wb.Update(CampaignFollower)
+		// blocks until elected or error
+		err := c.election.Campaign(ctx, c.val)
+		if err != nil {
+			c.wb.Update(err)
+		}
+
+		select {
+		case <-c.session.Done():
+			c.wb.Update(ErrSessionExpired)
```
cw9: Probably easier for user to use if you create a type
```
CampaignResponse struct {
  State CampaignState
  Err error
}
```
and set the State to CampaignError, Err = ErrSessionExpired for cases like this

services/leader/multierr.go
```diff
@@ -0,0 +1,24 @@
+package leader
+
+import "strings"
+
+type multiErr struct {
+	errs []error
+}
+
+func newMultiError(errs ...error) error {
```
cw9: could reuse MultiError in m3x/errors

services/leader/client.go
```diff
@@ -0,0 +1,229 @@
...
+	// will stop the campaign
+	c.cancelWithLock()
+
+	// TODO(mschalle): resign context timeout?
+	if err := c.election.Resign(context.Background()); err != nil {
+		c.wb.Update(err)
+		return err
+	}
+
+	c.wb.Update(CampaignFollower)
```
schallert: I would argue so. If you're not the leader but you haven't closed your client then follower sounds appropriate. I'll change the default state from `Unstarted` to `Follower` to reflect this.

services/leader/client.go
schallert: `Close()` errors if there's an error requesting that etcd revokes the lease. This could be because quorum has been lost, the lease was already expired, or a variety of other things. I think it's better to mark the client as closed if there's some weird state with etcd then assume we're still the leader.

services/leader/client_test.go
schallert: It seems as if a nil error is returned, I've added a test for this

schallert:
Still working on addressing other parts of feedback but have pushed options-related changes for now. The rest is coming shortly

services/leader/service.go
```diff
@@ -0,0 +1,175 @@
...
+	s.RLock()
+	errC := make(chan error, len(s.clients))
+	var wg sync.WaitGroup
+
+	for _, cl := range s.clients {
+		wg.Add(1)
+
+		go func(cl *client) {
+			if err := cl.Close(); err != nil {
+				errC <- err
```
schallert: I wanted to leave `multierr` as pretty much a convenience wrapper around an array of errors that's declared once rather than one that's declared and built up over time (similar to [go-multierror](https://github.com/hashicorp/go-multierror), which we could use instead if we wanted)

services/leader/client.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// Close closes the election service client entirely. No more campaigns can be
+// started and any outstanding campaigns are closed.
+func (c *client) close() error {
+	if c.setClosed() {
+		c.Lock()
+		c.cancelWithLock()
+		c.Unlock()
+
```
cw9: do we need to update c.campaigning here?

services/leader/client.go
```diff
@@ -0,0 +1,239 @@
...
+			c.wb.Update(okCampaignStatus(CampaignLeader))
+		}
+
+		cancel()
+		c.Lock()
+		c.ctxCancel = nil
+		c.Unlock()
+	}()
+
+	return c.w, nil
```
cw9: should create a new watch from the watchable and return it, otherwise if the watch is shared by multiple users and one of them closed it, other users will be affected

services/leader/client.go
```diff
@@ -0,0 +1,239 @@
...
+	return c.opts.DefaultHostname()
+}
+
+func (c *client) campaign(override string) (xwatch.Watch, error) {
+	if c.isClosed() {
+		return nil, errClientClosed
+	}
+
+	if !atomic.CompareAndSwapUint32(&c.campaigning, 0, 1) {
+		return nil, ErrCampaignInProgress
```
cw9: should we return a watch in this case? Just don't trigger a new campaign?

services/leader/client.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	c.Lock()
+	c.ctxCancel = cancel
+	c.Unlock()
+
+	go func() {
+		c.wb.Update(okCampaignStatus(CampaignFollower))
```
cw9: is this update necessary? I guess it's for recovering from err state?

services/leader/client.go
```diff
@@ -0,0 +1,239 @@
...
+		// blocks until elected or error
+		err := c.election.Campaign(ctx, c.val(override))
+		if err != nil {
+			c.wb.Update(errCampaignStatus(err))
```
cw9: Spoke offlne, we should probably handle all the error cases in the leader library. But we do want to notify users about the Error state

schallert:
Closing this PR to open another for cleaner history

PR #116, opened 2017-05-06
Allow user to specify namespace in KV

cw9:
This PR allows users to specify namespace when creating KV store as top level namespace to improve isolation between services.

This PR did not use the namespace feature in etcd because we need to upgrade to go 1.8 in order to use that, when we are ready to upgrade Go we can easily change the impl to use etcd namespace

@xichen2020 @jeromefroe  @dgromov

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -26,22 +26,22 @@ import (
 )
 
 // Client is the base interface into the cluster management system, providing
-// access to cluster services
+// access to cluster services.
 type Client interface {
-	// Services returns access to the set of services
+	// Services returns access to the set of services.
 	Services() (services.Services, error)
 
-	// KV returns access to the distributed configuration store
-	// To be deprecated
+	// KV returns access to the distributed configuration store.
+	// To be deprecated.
 	KV() (kv.Store, error)
 
-	// Txn returns access to the transaction store
-	// To be deprecated
+	// Txn returns access to the transaction store.
+	// To be deprecated.
 	Txn() (kv.TxnStore, error)
 
-	// Store returns access to the distributed configuration store with a namespace
+	// Store returns access to the distributed configuration store with a namespace.
 	Store(namespace string) (kv.Store, error)
 
-	// TxnStore returns access to the transaction store with a namespace
+	// TxnStore returns access to the transaction store with a namespace.
 	TxnStore(namespace string) (kv.TxnStore, error)
 }
--- a/client/client.go
+++ b/client/client.go
@@ -32,8 +32,16 @@ type Client interface {
 	Services() (services.Services, error)
 
 	// KV returns access to the distributed configuration store
+	// To be deprecated
 	KV() (kv.Store, error)
 
 	// Txn returns access to the transaction store
+	// To be deprecated
 	Txn() (kv.TxnStore, error)
+
+	// Store returns access to the distributed configuration store with a namespace
+	Store(namespace string) (kv.Store, error)
+
+	// TxnStore returns access to the transaction store with a namespace
+	TxnStore(namespace string) (kv.TxnStore, error)
 }
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -42,7 +42,7 @@ import (
 
 const (
 	internalPrefix     = "_"
-	cacheFileSeparater = "_"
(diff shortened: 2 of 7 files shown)


client/client.go
```diff
@@ -32,8 +32,16 @@ type Client interface {
 	Services() (services.Services, error)
 
 	// KV returns access to the distributed configuration store
+	// To be deprecated
```
xichen2020: Put a period at the end of the first line perhaps? Same goes for the rest

client/etcd/client.go
```diff
@@ -40,11 +41,15 @@ import (
 )
 
 const (
-	keySeparator    = "/"
-	cacheFileFormat = "%s-%s.json"
-	kvPrefix        = "_kv/"
+	internalPrefix     = "_"
+	cacheFileSeparater = "_"
```
xichen2020: `Separator`

client/etcd/client.go
```diff
@@ -199,27 +220,33 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
...
-	return filepath.Join(cacheDir, fileName(service, zone))
-}
-
-func fileName(service, zone string) string {
-	cacheFileName := fmt.Sprintf(cacheFileFormat, service, zone)
-
-	return strings.Replace(cacheFileName, string(os.PathSeparator), "_", -1)
+func (c *csclient) cacheFileFn(zone string) etcdkv.CacheFileFn {
+	return etcdkv.CacheFileFn(func(namespace string) string {
+		if c.opts.CacheDir() == "" || c.opts.Service() == "" {
```
xichen2020: Do we allow service to be empty in options?

client/etcd/client.go
```diff
@@ -199,27 +220,33 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
...
-	if env != "" {
-		res = concat(res, concat(env, keySeparator))
+func fileName(parts ...string) string {
+	// get non-empty parts
+	idx := 0
+	for _, part := range parts {
+		if part == "" {
+			continue
+		}
+		parts[idx] = part
```
xichen2020: nit: could compare the part index with `idx` here and don't set it if its' the same

client/etcd/client.go
```diff
@@ -124,24 +133,36 @@ func (c *csclient) createServices() {
...
+		SetInstrumentsOptions(instrument.NewOptions().
+			SetLogger(c.logger).
+			SetMetricsScope(c.kvScope)).
+		SetCacheFileFn(c.cacheFileFn(zone))
+
+	for _, namespace := range namespaces {
+		if namespace == "" {
+			continue
+		}
+		opts = opts.SetPrefix(opts.ApplyPrefix(namespace))
```
xichen2020: Doesn't this mean the prefixes will be structured in reverse order? Seems counter intuitive

cw9: no, it will be in the same order as the namespaces passed in, pretty much just keep appending namespaces

client/etcd/client.go
cw9: nope, gonna remove this check on empty service

client/etcd/client.go
xichen2020: Ah yeah, misread it

client/etcd/client_test.go
```diff
@@ -26,142 +26,145 @@ import (
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/integration"
 	"github.com/m3db/m3cluster/services"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
xichen2020: A blank line between m3db imports and 3rd party imports?

kv/etcd/options.go
```diff
@@ -35,8 +35,12 @@ var (
 	defaultWatchChanResetInterval = 10 * time.Second
 	defaultWatchChanInitTimeout   = 10 * time.Second
 	defaultRetryOptions           = xretry.NewOptions().SetMaxRetries(5)
+	defaultCacheFileFn            = func(namespace string) string { return "" }
```
xichen2020: nit: can just do `func(string) string {return ""}`

PR #122, opened 2017-05-16
Optimize shard distribution

cw9:
This diff optimizes shard distribution in the placement algorithm, this issue was found during the investigation of a case that the weights of the candidate instances differs slightly (see TestMinorWeightDifference), but the original algorithm caused the node with lower weight to be underloaded, the difference in loads among the cluster is disproportional to the difference in weights when the weight difference is small.

This diff introduces a Optimize() function in PlacementHelper and tries to bring up the load closer to the target load. This function also leads to a few potential improvements in the algo which will be tried in future diffs.

@xichen2020 @dgromov

### DIFF
--- a/services/placement/algo/sharded_helper.go
+++ b/services/placement/algo/sharded_helper.go
@@ -243,7 +243,8 @@ func (ph *placementHelper) moveOneShardInState(from, to services.PlacementInstan
 }
 
 func (ph *placementHelper) moveShard(candidateShard shard.Shard, from, to services.PlacementInstance) bool {
-	if !ph.canAssignInstance(candidateShard.ID(), from, to) {
+	shardID := candidateShard.ID()
+	if !ph.canAssignInstance(shardID, from, to) {
 		return false
 	}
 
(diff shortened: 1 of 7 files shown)


services/placement/algo/sharded_helper.go
```diff
@@ -37,28 +37,38 @@ const (
 	nonEmptyOnly
 )
 
+type optimizeType int
+
+const (
+	// safe optimizes the load distribution without violatin
```
xichen2020: `violating`

services/placement/algo/sharded_helper.go
```diff
@@ -341,7 +375,15 @@ func (ph *placementHelper) PlaceShards(shards []shard.Shard, from services.Place
 	return nil
 }
 
-func (ph *placementHelper) returnInitializingShardsToSource(shardSet map[uint32]shard.Shard, from services.PlacementInstance) error {
+func (ph *placementHelper) returnInitializingShardsToSource(
+	shardSet map[uint32]shard.Shard,
+	from services.PlacementInstance,
+	candidaes []services.PlacementInstance,
```
xichen2020: `candidates`

services/placement/algo/sharded_helper.go
```diff
@@ -350,46 +392,50 @@ func (ph *placementHelper) returnInitializingShardsToSource(shardSet map[uint32]
 		if sourceID == "" {
 			continue
 		}
-		sourceInstance, ok := ph.instances[sourceID]
+		sourceInstance, ok := candidateMap[sourceID]
 		if !ok {
-			return fmt.Errorf("could not find sourceID %s in the placement", sourceID)
+			continue
```
xichen2020: Why are we not returning an error here?

services/placement/algo/sharded_helper.go
```diff
@@ -350,46 +392,50 @@ func (ph *placementHelper) returnInitializingShardsToSource(shardSet map[uint32]
...
-			otherRack = append(otherRack, instance)
-		}
+		otherRack = append(otherRack, instance)
 	}
 
-	instanceHeap := ph.BuildInstanceHeap(nonLeavingInstances(otherRack), true)
-
+	instanceHeap, err := ph.buildInstanceHeap(nonLeavingInstances(otherRack), true)
+	if err != nil {
+		return nil
```
xichen2020: should this be `return err`?

cw9: good catch

services/placement/algo/sharded_helper.go
cw9: Basically it won't happen as the placement were always validated before reaching here, but I guess won't hurt to keep returning errors here.

PR #123, opened 2017-05-16
placement no longer gets cloned for every shard in MarkShardAvaialble

dgromov:
The placement clone is now create once before going through all the
shards. This commit also moves a bunch of stuff into placement.go from
the algo module.

### DIFF
--- a/services/placement/algo/sharded_helper_test.go
+++ b/services/placement/algo/sharded_helper_test.go
@@ -191,56 +191,6 @@ func TestNonLeavingInstances(t *testing.T) {
 	assert.Equal(t, "i2", r[1].ID())
 }
 
-func TestMarkShard(t *testing.T) {
-	i1 := placement.NewEmptyInstance("i1", "", "", "endpoint", 1)
-	i1.Shards().Add(shard.NewShard(1).SetState(shard.Available))
-	i1.Shards().Add(shard.NewShard(2).SetState(shard.Available))
-
-	i2 := placement.NewEmptyInstance("i2", "", "", "endpoint", 1)
-	i2.Shards().Add(shard.NewShard(1).SetState(shard.Initializing).SetSourceID("i3"))
-	i2.Shards().Add(shard.NewShard(2).SetState(shard.Initializing).SetSourceID("i1"))
-	i2.Shards().Add(shard.NewShard(3).SetState(shard.Initializing).SetSourceID("i1"))
-
-	p := placement.NewPlacement().
-		SetInstances([]services.PlacementInstance{i1, i2}).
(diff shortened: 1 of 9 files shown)


services/placement/placement.go
```diff
@@ -316,3 +354,79 @@ func (s ByIDAscending) Less(i, j int) bool {
...
+	for i, instance := range list {
+		if instance.ID() == instanceID {
+			return append(list[:i], list[i+1:]...)
+		}
+	}
+	return list
+}
+
+// MarkShardAvailable marks the state of a shard to available
+func MarkShardAvailable(p services.ServicePlacement, instanceID string, shardID uint32) (services.ServicePlacement, error) {
```
cw9: nit: you could create a markShardAvailable which does everything MarkShardAvailable does now, and MarkShardAvailable just do a clone and then call markShardAvailable, so the markShardAvailable can be reused in MarkAllShardsAsAvailable as well

services/placement/algo/sharded_helper_test.go
```diff
@@ -191,54 +191,14 @@ func TestNonLeavingInstances(t *testing.T) {
...
-	_, err = MarkShardAvailable(p, "i2", 3)
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "does not exist in source instance")
-
-	_, err = MarkShardAvailable(p, "i2", 2)
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "not leaving instance")
-}
-
 func TestRemoveInstanceFromArray(t *testing.T) {
```
cw9: this test should be moved

dgromov: I did. Its in placement_test now

services/placement/algo/sharded_helper_test.go
```diff
@@ -191,54 +191,14 @@ func TestNonLeavingInstances(t *testing.T) {
...
 		placement.NewEmptyInstance("i2", "", "", "endpoint", 1),
 	}
 
-	assert.Equal(t, instances, removeInstance(instances, "not_exist"))
-	assert.Equal(t, []services.PlacementInstance{placement.NewEmptyInstance("i2", "", "", "endpoint", 1)}, removeInstance(instances, "i1"))
+	assert.Equal(t, instances, placement.RemoveInstance(instances, "not_exist"))
+	assert.Equal(t, []services.PlacementInstance{placement.NewEmptyInstance("i2", "", "", "endpoint", 1)}, placement.RemoveInstance(instances, "i1"))
 }
 
 func TestCopy(t *testing.T) {
```
cw9: this should be moved as well

PR #124, opened 2017-05-17
Add CutoverNanos property and InstancesForShard method for ServicePlacement

xichen2020:
cc @cw9 

This PR adds `CutoverNanos()` and `InstancesForShard` to the `ServicePlacement` interface so they can consumed by users outside m3cluster.


services/types.go
```diff
@@ -278,6 +286,15 @@ type ServicePlacement interface {
...
 	SetIsSharded(v bool) ServicePlacement
 
+	// CutoverNanos returns the cutover time in nanoseconds.
+	CutoverNanos() int64
+
+	// SetCutoverNanos sets the cutover time in nanoseconds.
+	SetCutoverNanos(cutoverNanos int64) ServicePlacement
+
+	// Proto generates a placement proto.
+	Proto() (*placementproto.Placement, error)
```
cw9: I'm not sure if it's a good idea to tie all these interfaces into all the specific proto definitions

xichen2020: I don't see how this is worse than the proto utility functions however?

services/placement/staged_placement.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return placement, nil
+}
+
+// activeIndexWithLock finds the index of the last placement whose cutover time is no
+// later than t (a.k.a. the active placement). The cutover times of the placements are
+// sorted in ascending order (i.e., earliest time first).
+func (p *activeStagedPlacement) activeIndexWithLock(timeNanos int64) int {
+	idx := 0
+	for idx < len(p.placements) && p.placements[idx].CutoverNanos() <= timeNanos {
```
cw9: iterate from len to 0?

services/types.go
```diff
@@ -248,6 +250,108 @@ type PlacementOptions interface {
...
+
+	// SetOnPlacementsRemovedFn sets the callback function for removing placement.
+	SetOnPlacementsRemovedFn(value OnPlacementsRemovedFn) ActiveStagedPlacementOptions
+
+	// OnPlacementsRemovedFn returns the callback function for removing placement.
+	OnPlacementsRemovedFn() OnPlacementsRemovedFn
+}
+
+// StagedPlacement describes a series of placements applied in staged fashion.
+type StagedPlacement interface {
```
cw9: you also need the function to generate Proto for this interface?

services/placement/staged_placement_watcher.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (t *stagedPlacementWatcher) process(value interface{}) error {
+	t.Lock()
+	defer t.Unlock()
+
+	if t.state != placementWatcherWatching {
+		return errPlacementWatcherIsNotWatching
+	}
+	ps := value.(services.StagedPlacement)
+	placement := ps.ActiveStagedPlacement(t.nowFn().UnixNano())
```
cw9: If the latest update from kv contains a new placement with a cutover time in the future, this watcher would only ref to the active placement at the time the update was received and keep returning this placement in ActiveStagedPlacement() even if the time passes the future cutover time?

services/types.go
```diff
@@ -243,18 +250,123 @@ type PlacementOptions interface {
...
+
+	// InitWatchTimeout returns the initial watch timeout.
+	InitWatchTimeout() time.Duration
+}
+
+// ActiveStagedPlacement describes active staged placement.
+type ActiveStagedPlacement interface {
+	// ActivePlacement returns the currently active placement for a given time, the callback
+	// function when the caller is done using the placement, and any errors encountered.
+	ActivePlacement() (ServicePlacement, DoneFn, error)
```
cw9: Can you rename ServicePlacement to Placement?

services/placement/staged_placement.go
xichen2020: iterate from 0 is going to be faster because in most cases the active placement is the first one, and if it isn't, the expire goroutine will make it so.

services/types.go
xichen2020: Sure.

services/types.go
xichen2020: Not for now, although I guess yeah if we want to generate a new placement and persist in kv, sure can add.

services/placement/staged_placement_watcher.go
xichen2020: ActiveStagedPlacement will return the currently active placement alongside all placements in the future. To get the current active placement, users need to call `.ActivePlacement()` on the active staged placement. This way if the time passes the future cutover time, the correct active placement will be chosen.

There is still an issue with clock skews, but I'll handle that in a separate diff when I start working on topology changes.

PR #125, opened 2017-05-26
Dgromov/use watch events

dgromov:
Watch update function will now use most recent event data instead of re-getting the value.

### DIFF
--- a/etcd/watchmanager/types.go
+++ b/etcd/watchmanager/types.go
@@ -35,7 +35,7 @@ type WatchManager interface {
 }
 
 // UpdateFn is called when an event on the watch channel happens
-type UpdateFn func(key string) error
+type UpdateFn func(key string, events []*clientv3.Event) error
 
 // TickAndStopFn is called every once a while
 // to check and stop the watch if needed
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -392,16 +392,18 @@ func (c *client) update(key string, events []*clientv3.Event) error {
 		lastEvent = events[len(events)-1]
 		nv = newValue(lastEvent.Kv.Value, lastEvent.Kv.Version, lastEvent.Kv.ModRevision)
 	}
-	defer c.mergeCache(key, nv.(*value))
+
+	if nv != nil {
+		defer c.mergeCache(key, nv.(*value))
+	}
 
 	if curValue == nil {
 		// At watch creation or at key creation, just update the watch to current value.
-		if noEvents || lastEvent.IsCreate() {
+		if noEvents || lastEvent.Type != clientv3.EventTypeDelete {
 			return w.Update(nv)
 		}
 	}
 
-	// now both curValue and newValue are valid, compare version
 	if nv.IsNewer(curValue) {
 		if lastEvent.Type == clientv3.EventTypeDelete {
 			return w.Update(nil)
--- a/kv/etcd/store.go
+++ b/kv/etcd/store.go
@@ -355,22 +355,28 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
 	return w, err
 }
 
-func (c *client) update(key string) error {
+func (c *client) update(key string, events []*clientv3.Event) error {
 	var (
-		newValue kv.Value
-		err      error
+		nv        kv.Value
+		err       error
+		lastEvent *clientv3.Event
 	)
-	// we need retry here because if Get() failed on an watch update,
-	// it has to wait 10 mins to be notified to try again
-	if execErr := c.retrier.Attempt(func() error {
-		newValue, err = c.get(key)
-		if err == kv.ErrNotFound {
-			// do not retry on ErrNotFound
-			return xretry.NonRetryableError(err)
+
+	noEvents := events == nil || len(events) == 0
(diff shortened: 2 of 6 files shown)


etcd/watchmanager/types.go
```diff
@@ -35,7 +35,7 @@ type WatchManager interface {
 }
 
 // UpdateFn is called when an event on the watch channel happens
-type UpdateFn func(key string) error
+type UpdateFn func(key string, events []*clientv3.Event) error
```
cw9: You could consider adding a new UpdateFn one with events and one without? would that make things clearer?

dgromov: It makes sense to have all UpdateFns to use the events instead of re-geting. I rather leave this and have the option to use it later.

cw9: But when there is no watch, there is no event you have to call get() and when there is a watch, there is no need to call get()?

cw9: Or you could handle the withEvents and noEvents path separately in the updateFn and only share the mergeCache/updateWatch part?

dgromov: Sure, I can split them up. I didn't want to do that initially because I wasn't sure to do with the case where the event slice was not nil but just empty. I guess we can just skip calling the updateFn and emit a warning or something in that case.

cw9: seems you are treating nil slice and empty slice the same right now? you can still keep this logic in updateFn? Also Do you know what's contained in the watch response when a progress watch update happens?

kv/etcd/store.go
```diff
@@ -355,22 +355,63 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
...
+	return nv, err
+}
+
+func (c *client) getFromEtcdEvents(key string, events []*clientv3.Event) (kv.Value, error) {
+	var (
+		nv        kv.Value
+		err       error
+		lastEvent *clientv3.Event
+	)
+	noEvents := events == nil || len(events) == 0
```
cw9: do this before defining those vars

kv/etcd/store.go
```diff
@@ -355,22 +355,63 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
...
+
+	if noEvents {
+		return nil, fmt.Errorf("unexpected: no events found for key: %s", key)
+	}
+
+	lastEvent = events[len(events)-1]
+	nv = newValue(lastEvent.Kv.Value, lastEvent.Kv.Version, lastEvent.Kv.ModRevision)
+	if lastEvent.Type == clientv3.EventTypeDelete {
```
cw9: do this before creating nv?

kv/etcd/store.go
cw9: actually seems you already checked this before calling getFromEtcdEvents?

kv/etcd/store.go
```diff
@@ -355,22 +355,63 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
...
 			// do not retry on ErrNotFound
 			return xretry.NonRetryableError(err)
 		}
 		return err
 	}); execErr != nil && xerrors.GetInnerNonRetryableError(execErr) != kv.ErrNotFound {
-		return execErr
+		return nil, execErr
+	}
+
+	return nv, err
```
cw9: hmm, interesting, should return nv, nil here? or ErrNotFound will be returned and cause the watches not getting deletion updates? This case should be covered in a test

kv/etcd/store.go
```diff
@@ -381,19 +422,19 @@ func (c *client) update(key string) error {
 	}
 
 	curValue := w.Get()
-	if curValue == nil && newValue == nil {
-		// nothing to update
+
+	// They have the same values, or both are nil. There is nothing to do.
+	if curValue == nv {
```
cw9: Would the 2 values ever have same values?

kv/etcd/store.go
dgromov: Maybe someone would want to use this function outside of the update call?

kv/etcd/store.go
dgromov: ErrNotFound gets handled in the return nil, execErr above but yeah, good call.

kv/etcd/store.go
dgromov: Probably not, I just figured it should get handled the same way if it ever is. I'll change it back to check for nils

kv/etcd/store.go
cw9: if err can only be nil here, it's clearer to just return nil

kv/etcd/store.go
```diff
@@ -355,22 +355,56 @@ func (c *client) Watch(key string) (kv.ValueWatch, error) {
...
 	}); execErr != nil && xerrors.GetInnerNonRetryableError(execErr) != kv.ErrNotFound {
-		return execErr
+		return nil, execErr
+	}
+
+	return nv, err
+}
+
+func (c *client) getFromEtcdEvents(key string, events []*clientv3.Event) kv.Value {
+	var (
```
cw9: nit: remove the vars and just do lastEvent := and nv := when they were created?

PR #127, opened 2017-06-06
[test] Kv/etcd: prevent TestWatchLastVersion race

schallert:
I've run into this race a few times on builds for other PRs (example:
https://travis-ci.org/m3db/m3cluster/builds/240110273#L1113)  and think
this should fix it (specifically the race between `require.NoError()`
and the testing object `t` itself, not necessarily the grpc error).

EDIT: Actually I think this should may the grpc error as well since all of
the `store.Set` calls should be fully finished before we start closing the
etcd server.

### DIFF
--- a/kv/etcd/store_test.go
+++ b/kv/etcd/store_test.go
@@ -23,6 +23,7 @@ package etcd
 import (
 	"fmt"
 	"io/ioutil"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kv/etcd/store_test.go
```diff
@@ -272,14 +273,28 @@ func TestWatchLastVersion(t *testing.T) {
...
+	var wg sync.WaitGroup
 	lastVersion := 100
+	errC := make(chan error, lastVersion)
+	wg.Add(lastVersion)
 	go func() {
 		for i := 1; i <= lastVersion; i++ {
 			_, err := store.Set("foo", genProto(fmt.Sprintf("bar%d", i)))
-			require.NoError(t, err)
+			if err != nil {
+				errC <- err
```
cw9: instead of asserting no err here, we should atomic inc the number of errs from all the set and compare
value.Version() == lastVersion - atomic.LoadInt32(numErrs) in the next for loop, otherwise this test will still be flaky since there will be errors from etcd cluster.

kv/etcd/store_test.go
```diff
@@ -272,18 +274,28 @@ func TestWatchLastVersion(t *testing.T) {
 	require.NoError(t, err)
 	require.Nil(t, w.Get())
 
+	var (
+		wg   sync.WaitGroup
```
cw9: I think you could get rid of the wg?

PR #131, opened 2017-06-29
[services] Leader: election functionality

schallert:
Still working on some package-level docs but the comments in `services/types.go`
should serve as a sufficient high-level overview.

Compared to [the previous](https://github.com/m3db/m3cluster/pull/115) iteration of this PR, this one notably uses channels
instead of `xwatch.Watch`s, and does not include automatic retries. The channel
change is due to the nature of the caller needing to process each state event in
order, which might not happen if the state changes between when the `watch.C()`
channel was notified and when the caller received the value from `watch.Get()`.
The removal of retries is potentially temporary, but at least for now it greatly
simplifies managing the state of an election. Furthermore different clients may
need different retry logic (some might not want to retry campaigns at all), and
until we have a better idea of what the different use cases are it's not
difficult for clients needing retries to do so in a loop and checking the value
of `service.Leader()` if they encounter an error to determine if they need to
retry.


services/leader/client.go
```diff
@@ -0,0 +1,203 @@
...
+	return fmt.Sprintf(
+		keyFormat,
+		leaderKeyPrefix,
+		fmt.Sprintf(keyFormat, env, sid.Name()))
+}
+
+func electionPrefix(sid services.ServiceID, electionID string) string {
+	eid := electionID
+	if eid == "" {
+		eid = defaultElectionSuffix
```
cw9: nit: rename to defaultElectionID?

services/leader/service.go
```diff
@@ -0,0 +1,185 @@
...
+	sync.RWMutex
+
+	closed     uint32
+	clients    map[string]*clientEntry
+	opts       Options
+	etcdClient *clientv3.Client
+}
+
+// clientEntry stores a cached client as well as the TTL it was created with so
+// that a user will receive and error if they try to create a new client with a
```
cw9: typo: receive "an" error?

services/leader/service.go
```diff
@@ -0,0 +1,185 @@
...
+func (s *multiClient) closeClients() error {
+	s.RLock()
+	errC := make(chan error, len(s.clients))
+	var wg sync.WaitGroup
+
+	for _, cl := range s.clients {
+		wg.Add(1)
+
+		go func(cl *client) {
+			if err := cl.close(); err != nil {
```
cw9: is cl.close() expected to take a long time?

services/leader/service.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	ce, ok = s.clients[electionID]
+	if ok {
+		// another client was created between RLock and now, close new one
+		go clientNew.close()
+		return ce.client, nil
```
cw9: ce.client might be created with a different ttl though, we should compare ttl here?

services/leader/service.go
```diff
@@ -0,0 +1,185 @@
...
+	client, err := s.getOrCreateClient(electionID, ttl)
+	if err != nil {
+		return nil, err
+	}
+
+	if opts == nil {
+		opts = services.NewCampaignOptions()
+	}
+
+	return client.campaign(opts.LeaderValue(), opts)
```
cw9: any reason not just campaign(opts)?

services/services.go
```diff
@@ -258,3 +260,91 @@ func (i PlacementInstances) String() string {
...
+func (e electionOpts) String() string {
+	return fmt.Sprintf("[defaultHostname: %s, leaderTimeout: %s, resignTimeout: %s]",
+		e.defaultHostname,
+		e.leaderTimeout.String(),
+		e.resignTimeout.String())
+}
+
+type campaignOpts struct {
+	val string
+	ro  xretry.Options
```
cw9: Not used?

services/types.go
```diff
@@ -454,3 +459,77 @@ type HeartbeatService interface {
...
+	// for advice on proper usage and common gotchas.
+	//
+	// The leader will announce its hostname to observers unless opts is non-nil
+	// and opts.LeaderValue() is non-empty.
+	//
+	// NOTE: Once a campaign for a given electionID has been started, if it is
+	// lost or resigned and restarted it will be bound to the same TTL as the
+	// original due to constraints of etcd TTL sessions. The same applies if a
+	// call to Campaign() is made after a call to Leader() with the same ttl.
+	Campaign(electionID string, ttl int, opts CampaignOptions) (<-chan campaign.Status, error)
```
cw9: Do we have a use case to campaign with different ttls under same LeaderService client? Would it be cleaner to move this ttl to ElectionOptions? so we don't need the logic to check different ttl for the same electionID etc.

services/types.go
```diff
@@ -454,3 +459,77 @@ type HeartbeatService interface {
...
+	// returned from etcd. Defaults to 30 seconds.
+	ResignTimeout() time.Duration
+	SetResignTimeout(t time.Duration) ElectionOptions
+
+	// DefaultHostname returns the value that will be placed in the key of the
+	// leader of the election iff (1) no CampaionOptions option is passed to
+	// Campaign() and (2) the call to os.Hostname() fails. This value will very
+	// rarely be used since os.Hostname() should be reliable, and defaults to
+	// "default_hostname".
+	DefaultHostname() string
```
cw9: rename to DefaultValue?

services/leader/service.go
schallert: It's entirely possible for it to. Under the hood it makes a `Revoke` API call to etcd for the underlying lease. In most cases ideally it'd finish quickly but I didn't want to serialize a potentially large number of remote API calls.

services/types.go
schallert: I forget the exact reasoning but I think after an offline discussion we explicitly switched it to being on a per-call basis based on [this commit](https://github.com/m3db/m3cluster/pull/115/commits/2ab194bab69087ae09b8e148a37212f42d5493e5#diff-0c81ea9cafb4f2abe29474bb32f0bd15) from the last PR. We can switch back if we want to

services/leader/service.go
```diff
@@ -0,0 +1,235 @@
...
+election is in LEADER state and the user calls Resign() it will transition back
+to FOLLOWER state. Finally, if an election is in FOLLOWER state and a blocking
+call to Campaign() is ongoing and the user calls Resign(), the campaign will be
+cancelled and the election back in FOLLOWER state.
+
+The leader package requires some care to be used correctly. Calls to Resign()
+should ideally be done from the same goroutine as the call to Campaign().
+Multiple simultaneous calls to Campaign() are not guaranteed to succeed.
+Specifically, while a call to Campaign() will update the election value if the
+caller is the leader, a second call to Campaign() while a first is blocking will
```
cw9: I forgot the reason why we ended up with this semantic, can we return error for that second Campaign() call?

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+
+	return sc, nil
+}
+
+func (c *client) resign() error {
+	if c.isClosed() {
+		return errClientClosed
+	}
+
+	defer c.resetCampaign()
```
cw9: should we resetCampaign if resign failed?

schallert: Fixed!

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, err
+	}
+
+	var sessionOpts []concurrency.SessionOption
+	if ttl != 0 {
+		sessionOpts = append(sessionOpts, concurrency.WithTTL(ttl))
+	}
+
+	pfx := electionPrefix(opts.ServiceID(), electionID)
+	ec, err := election.NewClient(cli, pfx, election.WithSessionOptions(sessionOpts...))
```
xichen2020: Doesn't look like you actually need a slice here?

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+// their TTLs refreshed: https://github.com/coreos/etcd/issues/2660
+
+type client struct {
+	sync.RWMutex
+
+	ecl         *election.Client
+	opts        services.ElectionOptions
+	ctxCancel   context.CancelFunc
+	resignC     chan struct{}
+	campaigning uint32
```
xichen2020: Any reason why we use atomic here? Don't think these are perf-sensitive and you already have a mutex, would be much easier to comprehend the logic if we simply use locks and change `campaigning` and `closed` to `bool`

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+	if c.isClosed() {
+		return nil, errClientClosed
+	}
+
+	if !c.startCampaign() {
+		return nil, ErrCampaignInProgress
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	c.Lock()
```
xichen2020: Looks like only one goroutine can get here, do we still need to lock?

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), c.opts.ResignTimeout())
+	defer cancel()
+	if err := c.ecl.Resign(ctx); err != nil {
+		return err
+	}
+
+	c.resetCampaign()
+
+	// if successfully resigned and there was a campaign in Leader state cancel
```
xichen2020: Should we move this before `c.resetCampaign()`? Otherwise I can think of a scenario where this goroutine resets the campaign, then another goroutines starts a new campaign and becomes the leader, then this goroutine pushes a signal to the `resignC` channel, then the other goroutine reads it off the channel and becomes a follower even though it shouldn't.

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+	c.Lock()
+	if c.ctxCancel != nil {
+		c.ctxCancel()
+		c.ctxCancel = nil
+	}
+	c.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), c.opts.ResignTimeout())
+	defer cancel()
+	if err := c.ecl.Resign(ctx); err != nil {
```
xichen2020: Just curious, what type of errors can be returned here?

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func (c *client) leader() (string, error) {
+	if c.isClosed() {
+		return "", errClientClosed
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), c.opts.LeaderTimeout())
+	ld, err := c.ecl.Leader(ctx)
+	cancel()
```
xichen2020: nit: `defer cancel()` after creating the context might be more idiomatic.

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+	if err == concurrency.ErrElectionNoLeader {
+		return ld, ErrNoLeader
+	}
+	return ld, err
+}
+
+func (c *client) startCampaign() bool {
+	return atomic.CompareAndSwapUint32(&c.campaigning, 0, 1)
```
xichen2020: Yeah might be better if we simply use `c.Lock` and `c.Unlock` here.

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func (c *client) isClosed() bool {
+	return atomic.LoadUint32(&c.closed) == 1
+}
+
+// val returns the value the leader should propose based on (1) a potentially
+// empty override value and (2) the hostname of the caller (with a fallback to
+// the DefaultValue option).
+func (c *client) val(override string) string {
```
xichen2020: What's the use case when the caller will pass in an override?

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return c.opts.Hostname()
+}
+
+// elections for a service "svc" in env "test" should be stored under
+// "_ld/test/svc". A service "svc" with no environment will be stored under
+// "_ld/svc".
+func servicePrefix(sid services.ServiceID) string {
+	env := sid.Environment()
```
xichen2020: Do we allow services with no environment?

services/leader/campaign/state.go
```diff
@@ -0,0 +1,43 @@
+// Package campaign encapsulates the state of a campaign.
```
xichen2020: Need license header.

services/leader/campaign/state_string.go
```diff
@@ -0,0 +1,16 @@
+// Code generated by "stringer -type State"; DO NOT EDIT
```
xichen2020: Need license header.

services/leader/client.go
```diff
@@ -0,0 +1,220 @@
+package leader
```
xichen2020: License header.

services/leader/doc.go
```diff
@@ -0,0 +1,41 @@
+/*
```
xichen2020: License header.

services/leader/options.go
```diff
@@ -0,0 +1,64 @@
+package leader
```
xichen2020: All files in the leader package would need license header.

services/leader/options.go
```diff
@@ -0,0 +1,64 @@
...
+func NewOptions() Options {
+	return options{
+		eo: services.NewElectionOptions(),
+	}
+}
+
+type options struct {
+	sid services.ServiceID
+	eo  services.ElectionOptions
+	val string
```
xichen2020: `val` and `ttl` aren't used.

services/leader/options.go
```diff
@@ -0,0 +1,64 @@
...
+	o.eo = eo
+	return o
+}
+
+func (o options) Validate() error {
+	if o.sid == nil {
+		return errors.New("leader options must specify service ID")
+	}
+
+	// this shouldn't happen since we have sane defaults but prevents user error
```
xichen2020: // This ...

Also comments should end with a period.

services/leader/options.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (o options) SetElectionOpts(eo services.ElectionOptions) Options {
+	o.eo = eo
+	return o
+}
+
+func (o options) Validate() error {
+	if o.sid == nil {
+		return errors.New("leader options must specify service ID")
```
xichen2020: Define them as file-scope vars?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+var (
+	// errClientClosed indicates the election service client has been closed and
+	// no more elections can be started.
+	errClientClosed = errors.New("election client is closed")
+)
+
+type multiClient struct {
+	sync.RWMutex
+
+	closed     uint32
```
xichen2020: Probably better to make this a `bool` and use the `RWMutex` to protect its access.

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	clients    map[string]*clientEntry
+	opts       Options
+	etcdClient *clientv3.Client
+}
+
+// clientEntry stores a cached client as well as the TTL it was created with so
+// that a user will receive an error if they try to create a new client with a
+// different TTL
```
xichen2020: nit: add a period at the end.

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	return nil
+}
+
+func (s *multiClient) isClosed() bool {
+	return atomic.LoadUint32(&s.closed) == 1
+}
+
+func (s *multiClient) closeClients() error {
+	s.RLock()
+	errC := make(chan error, len(s.clients))
```
xichen2020: The convention is to keep the first enqueued error because if a close fails, it's likely all are failing for the same reason.

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	var wg sync.WaitGroup
+
+	for _, cl := range s.clients {
+		wg.Add(1)
+
+		go func(cl *client) {
+			if err := cl.close(); err != nil {
```
xichen2020: Do these clients share the same underlying etcd client? If so, would it be okay for the etcd client to be closed multiple times from multiple goroutines?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+				errC <- err
+			}
+			wg.Done()
+		}(cl.client)
+	}
+
+	s.RUnlock()
+
+	wg.Wait()
```
xichen2020: Should probably enforce some kind of timeout here? Not sure what's the semantics of the `close` call here? Could it block indefinitely?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	close(errC)
+
+	merr := xerrors.NewMultiError()
+	for err := range errC {
+		if err != nil {
+			merr = merr.Add(err)
+		}
+	}
+
+	if merr.NumErrors() > 0 {
```
xichen2020: nit: can just be `return merr.FinalError()`

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	opts       Options
+	etcdClient *clientv3.Client
+}
+
+// clientEntry stores a cached client as well as the TTL it was created with so
+// that a user will receive an error if they try to create a new client with a
+// different TTL
+type clientEntry struct {
+	client *client
+	ttl    int
```
xichen2020: Should this be `time.Duration`? Or does it have a time unit?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (s *multiClient) getOrCreateClient(electionID string, ttl int) (*client, error) {
+	const errTTLFmt = "cannot create client with ttl=(%d), already created with ttl=(%d)"
+
+	s.RLock()
+	ce, ok := s.clients[electionID]
+	s.RUnlock()
+	if ok {
+		if ce.ttl != ttl {
```
xichen2020: Seems like the map should be keyed by the combination of election id and ttl?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+			return nil, fmt.Errorf(errTTLFmt, ttl, ce.ttl)
+		}
+		return ce.client, nil
+	}
+
+	clientNew, err := newClient(s.etcdClient, s.opts, electionID, ttl)
```
xichen2020: hm why not create the client inside the write lock so we don't need to close it again if it's created between releasing the rlock and acquiring the wlock?

services/leader/service.go
```diff
@@ -0,0 +1,189 @@
...
+	keyFormat       = "%s/%s"
+)
+
+var (
+	// errClientClosed indicates the election service client has been closed and
+	// no more elections can be started.
+	errClientClosed = errors.New("election client is closed")
+)
+
+type multiClient struct {
```
xichen2020: Would it be better to define a `Client` interface with the same methods defined for `multiClient` in the interface and have both `client` and `multiClient` implement that interface?

services/services.go
```diff
@@ -258,3 +259,88 @@ func (i PlacementInstances) String() string {
 	strs = append(strs, "]")
 	return strings.Join(strs, "")
 }
+
+// NewElectionOptions returns an empty ElectionOptions.
+func NewElectionOptions() ElectionOptions {
+	eo := electionOpts{
+		leaderTimeout: 30 * time.Second,
```
xichen2020: Define these default options as consts? e.g.,
```
const (
  defaultLeaderTimeout = 30 * time.Second
  ...
)
```
and then
```
eo := electionOpts{
  leaderTimeout: defaultLeaderTimeout
  ..
}
```

services/services.go
```diff
@@ -258,3 +259,88 @@ func (i PlacementInstances) String() string {
...
+		resignTimeout: 30 * time.Second,
+		defaultValue:  "default_hostname",
```
xichen2020: Should the default value be the host name then?

services/types.go
```diff
@@ -454,3 +459,77 @@ type HeartbeatService interface {
...
+
+	// DefaultValue returns the value that will be placed in the key of the
+	// leader of the election iff (1) no CampaionOptions option is passed to
+	// Campaign() and (2) the call to os.Hostname() fails. This value will very
+	// rarely be used since os.Hostname() should be reliable, and defaults to
+	// "default_hostname".
+	DefaultValue() string
+	SetDefaultValue(s string) ElectionOptions
+
+	// Hostname returns the hostname of the host if accessible, otherwise the
```
xichen2020: Why is the host name part of the options?

services/types.go
```diff
@@ -454,3 +459,77 @@ type HeartbeatService interface {
...
+	// value for DefaultValue().
+	Hostname() string
+
+	String() string
```
xichen2020: Why do we need a string method for the options?

services/types.go
```diff
@@ -454,3 +459,77 @@ type HeartbeatService interface {
...
+
+// CampaignOptions provide the ability to override campaign defaults.
+type CampaignOptions interface {
+	// LeaderValue allows the user to override the value a campaign announces
+	// (that is, the value an observer sees upon calling Leader()). This defaults to
+	// the hostname of the caller.
+	LeaderValue() string
+	SetLeaderValue(v string) CampaignOptions
+
+	String() string
```
xichen2020: Same, seems a bit weird to define a string method for options.

services/client/options.go
```diff
@@ -101,6 +113,10 @@ func (o options) Validate() error {
 		return errNoHeartbeatGen
 	}
 
+	if o.ldGen == nil {
```
schallert: I think that at least for now it follows the constraints we've set, i.e. we require that a heartbeat gen function be set even though some service clients may only care about kv

services/client/services.go
```diff
@@ -542,6 +579,10 @@ func validateAdvertisement(sid services.ServiceID, id string) error {
 	return nil
 }
 
+func leaderCacheKey(sid services.ServiceID, opts services.ElectionOptions) string {
+	return fmt.Sprintf("%s/%s", sid.String(), opts.String())
```
schallert: We don't want it to be unique though, we just want to make sure that if `client.LeaderService(sid, opts)` is called twice with the same `sid` and `opts` that we return the already created `LeaderService` instance

services/leader/service.go
schallert: I think it makes more sense to keep it as an `int` since that's the type we pass to [`concurrency.WithTTL`](https://godoc.org/github.com/coreos/etcd/clientv3/concurrency#WithTTL).

services/leader/client.go
schallert: The only errors that can be returned are from us if the client was already closed, or from etcd if the API call to `Delete(leaderKey)` fails for any reason

services/leader/client.go
schallert: I thought we need it since multiple goroutines can simultaneously read in `resign()`?

services/leader/client.go
schallert: Removing this, it was an artifact from before I put the logic in `CampaignOptions` itself

schallert: Any time they want to announce a leader value that's not their hostname

services/leader/client.go
schallert: I think so? We treat this similarly in `services/client/util.go`:
```go
func serviceKey(s services.ServiceID) string {
	if s.Environment() == "" {
		return s.Name()
	}
	return fmt.Sprintf(keyFormat, s.Environment(), s.Name())
}
```

services/leader/service.go
schallert: While this does close our `leader.client` under the hood it doesn't actually close the underlying etcd client, just the session associated with the election.

services/leader/service.go
schallert: It *potentially* could, since it requires issuing a `Delete` call to etcd. I'll add a timeout

services/leader/service.go
schallert: Since `newClient` requires a remote API call to etcd I didn't want to block inside the write lock for the duration of that call

services/services.go
schallert: This is the value that's fallen back to if `os.Hostname()` returns an error

services/types.go
schallert: Only used for cache keys in `services/client/services.go`

services/leader/service.go
schallert: What would be the benefit of doing this? We only end up exposing `multiClient` anyway

services/leader/service.go
schallert: We only want a client with a single TTL per `electionID` though

services/client/options.go
xichen2020: Yeah that's fine.

services/client/services.go
xichen2020: Okay in that case I'd rename `String` to `ID` on the options just to be more explicit.

services/leader/client.go
xichen2020: Yeah fair enough.

services/leader/client.go
xichen2020: Sounds good then.

services/leader/service.go
xichen2020: I initially though the `multiClient` is exposed as a struct, which would make it hard to mock out the client behavior outside the package. However it seems we are using it as the `LeaderService` interface, so that should fine.

services/leader/service.go
xichen2020: kk makes sense.

services/leader/service.go
xichen2020: I feel that's probably okay given the creating the client is a low qps operation. This way is fine but feels kinda weird that you need to close out the resource if someone else got there ahead of you.

services/types.go
xichen2020: As discussed above probably makes sense to rename it to `ID` to be more explicit about its purpose.

services/client/services_test.go
```diff
@@ -956,6 +967,49 @@ func TestCacheCollisions_Watchables(t *testing.T) {
...
+			_, err := cl.LeaderService(sid, eo)
+			assert.NoError(t, err)
+		}
+	}
+
+	assert.Equal(t, 6, len(cl.(*client).ldSvcs),
+		"should cache 6 unique client entries")
+}
+
+func newTestLeaderGen(mc *gomock.Controller) LeaderGen {
```
cw9: do you need the gomock here? doesn't look like it's used?

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+// THE SOFTWARE.
+
+package leader
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/coreos/etcd/clientv3"
```
cw9: more etcd imports to the external section in the bottom?

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+	// buffer 1 to not block initial follower update
+	sc := make(chan campaign.Status, 1)
+
+	sc <- campaign.NewStatus(campaign.Follower)
+
+	go func() {
+		defer func() {
+			close(sc)
+			cancel()
+			c.resetCampaign()
```
xichen2020: nit: `c.stopCampaign` might be a better name?

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+	c.Lock()
+	c.campaigning = false
+	c.Unlock()
+}
+
+// Close closes the election service client entirely. No more campaigns can be
+// started and any outstanding campaigns are closed.
+func (c *client) close() error {
+	c.Lock()
+	c.closed = true
```
xichen2020: Might wanna check if `c.closed` is already `true` here?

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+// Appended to elections with an empty string for electionID to make it easier
+// for user to debug etcd keys.
+const defaultElectionID = "default"
+
+var (
+	// ErrNoLeader is returned when a call to Leader() is made to an election
+	// with no leader. We duplicate this error so the user doesn't have to
+	// import etcd's concurrency package in order to check the cause of the
+	// error.
+	ErrNoLeader = concurrency.ErrElectionNoLeader
```
cw9: nit: might want to create ErrNoLeader with a different error msg here, otherwise if this error is logged and if you search by the error msg, you can only get to the etcd package where concurrency.ErrElectionNoLeader is defined, might be hard to debug.

services/leader/client.go
xichen2020: Can we perhaps just always use `override` (which seems to come from `opts.LeaderValue()`) and remove this function?

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	return fmt.Sprintf(
+		keyFormat,
+		leaderKeyPrefix,
+		fmt.Sprintf(keyFormat, env, sid.Name()))
+}
+
+func electionPrefix(sid services.ServiceID, electionID string) string {
+	eid := electionID
```
xichen2020: nit: perhaps we should just always use `electionID` and get rid of `defaultElectionID`

services/leader/service.go
```diff
@@ -0,0 +1,194 @@
...
+	if s.isClosed() {
+		return nil, errClientClosed
+	}
+
+	client, err := s.getOrCreateClient(electionID)
+	if err != nil {
+		return nil, err
+	}
+
+	if opts == nil {
```
xichen2020: Probably don't want to do that here and instead fail loudly.

services/leader/client.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// if successfully resigned and there was a campaign in Leader state cancel
+	// it
+	select {
+	case c.resignCh <- struct{}{}:
+	default:
+	}
+
+	c.resetCampaign()
```
cw9: does resetCampaign need to be called in both here and in the go routine in campaign()?

services/leader/client.go
cw9: Yes we allow, but we never really do that in production, I'm considering always enforce a non-nil environment everywhere, would make logic like this cleaner.

services/services.go
```diff
@@ -31,6 +32,11 @@ import (
 	"github.com/m3db/m3cluster/shard"
 )
 
+const (
+	defaultLeaderTimeout = 30 * time.Second
```
xichen2020: hm that's a really long timeout, perhaps change it to something like 10 seconds? Same goes for the resign timeout.

services/services.go
```diff
@@ -258,3 +264,75 @@ func (i PlacementInstances) String() string {
...
+	}
+	eo.hostname = h
+
+	return eo, nil
+}
+
+type electionOpts struct {
+	leaderTimeout time.Duration
+	resignTimeout time.Duration
+	ttl           int
```
xichen2020: `ttlSecs` is probably a better name.

services/leader/service.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (s *multiClient) isClosed() bool {
+	s.RLock()
+	defer s.RUnlock()
+	return s.closed
+}
+
+func (s *multiClient) closeClients() error {
+	s.RLock()
+	errC := make(chan error, 1)
```
cw9: can we use a multierror instead of the error channel here?

services/services.go
```diff
@@ -258,3 +264,75 @@ func (i PlacementInstances) String() string {
...
+func (e electionOpts) ResignTimeout() time.Duration {
+	return e.resignTimeout
+}
+
+func (e electionOpts) SetResignTimeout(t time.Duration) ElectionOptions {
+	e.resignTimeout = t
+	return e
+}
+
+func (e electionOpts) TTL() int {
```
xichen2020: Rename this to `TTLSecs` or `TTLSeconds`?

services/services.go
```diff
@@ -258,3 +264,75 @@ func (i PlacementInstances) String() string {
...
+	return e.ttl
+}
+
+func (e electionOpts) SetTTL(ttl int) ElectionOptions {
```
xichen2020: `SetTTLSecs`?

services/services.go
```diff
@@ -258,3 +264,75 @@ func (i PlacementInstances) String() string {
...
+	eo.hostname = h
+
+	return eo, nil
+}
+
+type electionOpts struct {
+	leaderTimeout time.Duration
+	resignTimeout time.Duration
+	ttl           int
+	hostname      string
```
xichen2020: Perhaps we should just call this `leaderValue`, initialize it to the hostname by default, and change `HostName` to `LeaderValue`?

services/leader/client.go
schallert: I think it's cleaner in terms of the keys in etcd if we don't allow empty election IDs, since the keys would look like `<prefix>//<leaseID>`

services/services.go
schallert: I think that conflicts too much with the `LeaderValue` in `CampaignOptions`, and that they should remain separate since these options are service-wide and campaign opts are per-campaign

services/client/services_test.go
schallert: Yeah I just needed something that can return a valid non-nil implementation of `services.LeaderService`

services/leader/client.go
schallert: Yeah, since if the caller doesn't explicitly resign but lose their session for some reason we want them to be able to re-campaign

services/leader/service.go
schallert: I had this but @xichen2020 said the convention is to just keep the first error: https://github.com/m3db/m3cluster/pull/131#discussion_r127121384

services/leader/client.go
```diff
@@ -0,0 +1,245 @@
...
+	if !c.startCampaign() {
+		return nil, ErrCampaignInProgress
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	c.Lock()
+	c.cancelFn = cancel
+	c.Unlock()
+
+	proposeVal := opts.LeaderValue()
```
xichen2020: Can we just use `LeaderValue` as is and get rid of `HostName()` from the options interface? We can make `LeaderValue` default to the hostname if needed.

services/leader/client.go
```diff
@@ -0,0 +1,240 @@
...
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/m3db/m3cluster/services"
+	"github.com/m3db/m3cluster/services/leader/campaign"
+	"github.com/m3db/m3cluster/services/leader/election"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/clientv3/concurrency"
```
robskillington: Hey this is strange that `etcd` is imported directly into what should be "backing KV db" agnostic code

services/leader/client.go
```diff
@@ -0,0 +1,240 @@
...
+	client      *election.Client
+	opts        services.ElectionOptions
+	cancelFn    context.CancelFunc
+	resignCh    chan struct{}
+	campaigning bool
+	closed      bool
+}
+
+// newClient returns an instance of an client bound to a single election.
+func newClient(cli *clientv3.Client, opts Options, electionID string) (*client, error) {
```
robskillington: This really should be in a package named `leader/etcd` or something like that, it requires an etcd client.

PR #137, opened 2017-07-27
Add placement algo for mirror mode

cw9:
The basic idea of the mirror algo is to group instances with same shardSetID into one virtual instance and apply the sharded algo.
This algo does not support add replica at this moment, added a TODO to support it in the future.
This algo supports:
- add x * rf instances, x can be any number, the x * rf new instances will be grouped into x virtual instance and the instances from the original placement will be grouped in to vitual nodes as well and the x new virtual instances will be added to the existing virtual instances. After moving shards on the virtual instances, the result shard distribution will be applied back to original instances and generate a new placement.
- Remove x * rf instances, similar to Add above.
- Replace 1 instance with 1 instance, the shards from the leaving instance will be completely moved to the adding instance.

After this change, some further changes are still needed on placement service or m3admin side.
1, m3admin reads clusto and adjust weights, this algorithm assumes the instances with the same shard set id have the same weight, there are edge cases in clusto where some hosts have a memory of 2xx001MB RAM, and some hosts have 2xx002MB RAM, it's up to the m3admin to decide what their weight should be.
2, m3admin or placement service decides how to pick and group available instances into groups with the same shard set id. The placement algorithm does not assign shard set id, it takes in instances with already fixed weight and shard set id. 
3, This algorithm also does not care about host or endpoint of the instance, it respects them and takes them as is. The user(m3admin/placementservice) has to decide how to group instances into the same shard set id, the algorithm does not care if they are the same port or not, it only assigns shards based on weight and shard set id, but in real production we may want to group the instances from different hosts with same port together with the same shard set id?

@xichen2020


xichen2020:
A couple high-level questions:
* What happens if we try to replace an instance with a new instance that has more / less memory than the other nodes in the group?
* I assume rack constraint is still honored as before?

Regarding the TODOs:
* I think that assumption in general is okay, but what happens if say we only have node with more/less memory? We'll just take the weight of the nodes in the same group as the new weight?
* The placement service needs to respect the rack constraint as before, and also has additional constraints like no two replicas of the same shard set id can live on the same host?
* It'll probably make things easier to debug if we make instances with the same port in the same group, but that's not a hard requirement.

cw9:
> What happens if we try to replace an instance with a new instance that has more / less memory than the other nodes in the group?

Right now the memory of the new instance will not be respected when we are replacing since it has to be a full replacement, I'm thinking maybe we should do a validation in the algo to return error if the weight/shardsetid does not match between the old and new instance, which means the m3admin tool needs to set the new instance with the correct weight/shardsetid(same value from old instance). 

> I assume rack constraint is still honored as before?

Yes


> I think that assumption in general is okay, but what happens if say we only have node with more/less memory? We'll just take the weight of the nodes in the same group as the new weight?

Not sure which TODO is this referring to, but the algo needs to take in nodes with same weights for a shardsetid, if nodes in same shardsetid have different weights, the algo will report error, so m3admin needs to set the weight correctly(do whatever custom logic to set/adjust weight)

> The placement service needs to respect the rack constraint as before, and also has additional constraints like no two replicas of the same shard set id can live on the same host?

Rack constraint is respected, the shardSetID will be used as the rack for the virtual instance, but there is no constraints like "no two replicas of the same shard set id can live on the same host" in the algo. Right now I havn't added the idea of "Host" to the algorithm, they are all just instances, I'm thinking maybe we handle this custom logic in the m3admin tool, or we could potentially add Host to the placement interface.

> It'll probably make things easier to debug if we make instances with the same port in the same group, but that's not a hard requirement.

True, but the algo right now is not erroring out if instances with different ports is in the same group, it does not really look at the port/endpoint info for each instance, it only cares about the ID/rack/shardSetID/Weight of the instances, if we want to put instances with the same port in the same group, this could be done in m3admin/placementService

xichen2020:
I see, yeah would prefer to have the logic that handles assigning shard set ids to raw instances live in the OSS and keep custom logic in the internal m3admin code to a minimum level.

Right now the placement service/algorithm assumes a one-to-one mapping between an instance and a host and doesn't handle scenarios where you may have multiple instances per host, which is why we need to mimic a host with a virtual instance. I wonder if it's worth handling the multi-instance-per-host situation more generally, since that can be useful not just for the mirrored algorithm, but for the sharded algorithm as well.

cw9:
The multi-instance-per-host situation is something that's kinda specific to the m3agg use case, you will never have this issue in sharded algorithm because all the instances from the same host will be having the same rack, and they won't own the same shard more than once due to rack-conflict, needless to check the host.

I'm gonna make a different pr in placement service to pair up hosts/instances and set shardsetid for initializing placement and adding instances, but that diff should have nothing to do with the placement algorithm, so it should not block the code review for this pr.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, err
+	}
+
+	mirror := placement.NewPlacement().
+		SetInstances(mirrorInstances).
+		SetReplicaFactor(1).
+		SetShards(p.Shards()).
+		SetIsSharded(true).
+		SetIsMirrored(true)
+	// NB(cw): Always clean up the shard states from previous placement changes,
```
xichen2020: We should look into whether this should be done inside the placement algorithm or from the aggregator when the topology change finishes.

services/placement/placement.go
```diff
@@ -112,9 +115,10 @@ func (p *placement) SetInstances(instances []services.PlacementInstance) service
 		sort.Sort(ByIDAscending(instances))
 	}
 
-	p.instancesByShard = instancesByShard
-	p.instances = instancesMap
-	return p
+	newp := *p
```
xichen2020: Hm why the change of semantics? If we do decide to do this, we need to find all the places where these APIs are used and make sure we replace `p.SetXXX(...)` with `p = p.SetXXX(...)`. Sounds like a lot of work so prefer we don't do it unless we have a good reason to do so.

services/placement/service/placementservice.go
```diff
@@ -350,7 +330,7 @@ func (ps placementService) findAddingInstance(
 	// if there is a rack not in the current placement, prefer that rack
 	for r, instances := range candidateRackMap {
 		if _, exist := placementRackMap[r]; !exist {
-			return instances[0], nil
+			return instances[:1], nil
```
xichen2020: Do we know `instances` has a size of at least 1?

services/client/storage.go
```diff
@@ -131,7 +131,7 @@ func (s *client) Placement(sid services.ServiceID) (services.Placement, int, err
 
 	p, err := placement.NewPlacementFromProto(&placementProto)
 	if p != nil {
-		p.SetVersion(v.Version())
+		p = p.SetVersion(v.Version())
```
cw9: This is not a bug, I changed the semantic for the setters on placement, now every setter will create a new placement, will gradually change all the setters in m3cluster to that semantic.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,310 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
cw9: Sure

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+	instances []services.PlacementInstance,
+	shards []uint32,
+	rf int,
+) (services.Placement, error) {
+	mirrorInstances, err := groupInstancesByShardSetID(instances, rf)
+	if err != nil {
+		return nil, err
+	}
+
+	mirrorPlacement, err := a.shardedAlgo.InitialPlacement(mirrorInstances, shards, 1)
```
cw9: sure

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, instance := range mirrorInstances {
+		ph, leavingInstance, err := newRemoveInstanceHelper(mirrorPlacement, instance.ID(), a.opts)
+		if err != nil {
+			return nil, err
+		}
+		// place the shards from the leaving instance to the rest of the cluster
```
cw9: sg

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		if err := ph.PlaceShards(leavingInstance.Shards().All(), leavingInstance, ph.Instances()); err != nil {
+			return nil, err
+		}
+
+		if mirrorPlacement, _, err = addInstanceToPlacement(ph.GeneratePlacement(nonEmptyOnly), leavingInstance, false); err != nil {
```
cw9: it means empty instance will not be added to the placement.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+	}
+	return placementFromMirror(mirrorPlacement, nonLeavingInstances(p.Instances()), p.ReplicaFactor())
+}
+
+func (a mirroredAlgorithm) AddInstances(
+	p services.Placement,
+	addingInstances []services.PlacementInstance,
+) (services.Placement, error) {
+	if !p.IsMirrored() {
```
cw9: sounds good.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, err
+	}
+	mirrorInstances, err := groupInstancesByShardSetID(addingInstances, p.ReplicaFactor())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, instance := range mirrorInstances {
+		if _, ok := mirrorPlacement.Instance(instance.ID()); ok {
+			return nil, fmt.Errorf("shard set id %s already exist in current placement", instance.ShardSetID())
```
cw9: the id of the virtual instance is the shard set id

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+		ph := newAddInstanceHelper(mirrorPlacement, instance, a.opts)
+		ph.AddInstance(instance)
+		mirrorPlacement = ph.GeneratePlacement(nonEmptyOnly)
+	}
+
+	return placementFromMirror(mirrorPlacement, nonLeavingInstances(append(p.Instances(), addingInstances...)), p.ReplicaFactor())
```
cw9: internally the placement contains a map of instances, p.Instances() will create a new slice each time so this should probably be fine.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+	p services.Placement,
+	addingInstances []services.PlacementInstance,
+) (services.Placement, error) {
+	if !p.IsMirrored() {
+		return nil, errMirrorAlgoOnNotMirroredPlacement
+	}
+
+	for _, instance := range addingInstances {
+		if _, exist := p.Instance(instance.ID()); exist {
+			// TODO(cw): Should we allow adding a leaving instance?
```
cw9: sounds good

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+	if !ok {
+		return nil, fmt.Errorf("instance %s not found in placement", instanceID)
+	}
+
+	shards := leavingInstance.Shards()
+	for _, s := range shards.All() {
+		// NB(cw): Mark Available as Initializing in case we use shard state
+		// to indicate cutover/cutoff time for the shard.
+		if s.State() == shard.Available {
+			s.SetState(shard.Initializing)
```
cw9: This needs to be Initializing because the whole shard set needs to be moved to the new instance.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+	}
+
+	addingInstance = addingInstance.SetShards(shards).SetShardSetID(leavingInstance.ShardSetID())
+
+	// NB(cw): Replace is supposed to be used when the leaving node is considered dead,
```
cw9: I was trying to keep the SourceID in each initializing shard to be the source shardSetID rather than instanceID, so the shards on each instance in the same shardSet are pretty much the same and only modify capacity per shardset. 
But since we are going to support more customized operations I'm going to make the source ids to be instanceIDs, in which case I can just reuse the sharded algorithm's replace() here.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+	)
+	for _, instance := range instances {
+		ssID := instance.ShardSetID()
+		weight := instance.Weight()
+		shardSetCount[ssID] = shardSetCount[ssID] + 1
+		if w, ok := shardSetMap[ssID]; ok {
+			if weight == w {
+				continue
+			}
+			return nil, fmt.Errorf("found different weights: %d and %d, for shardset id %s", w, weight, ssID)
```
cw9: Don't think it's too helpful since we don't know the id of the first instance, the debugging should be pretty easy with the weight and shardset info.

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		shardSetMap[ssID] = instance.Weight()
+		res = append(
```
cw9: yes

services/placement/algo/mirrored.go
```diff
@@ -0,0 +1,299 @@
...
+			res,
+			placement.NewInstance().
+				SetID(ssID).
+				SetRack(ssID).
```
cw9: that's placementservice's job to not allow that to happen, which will be done in a follow up diff.

services/placement/placement.go
cw9: There is not a lot of places for that kind of replacement, I usually only call the setters when returning a placement in the end of a capacity change. I was planning to gradually change all semantics to that way since it's used in most other packages, but don't feel strongly either way. I'm gonna revert this.

services/placement/service/placementservice.go
cw9: Yes

services/placement/algo/mirrored.go
cw9: This could be called in aggregator when topology change finishes, but this has to happen before the algorithm making any topology change because otherwise some shardSetID may get more instances than RF.
For example when you do a single instance replace, that shardSetID will have the old leaving instance and the extra new instance, which would cause problems in the mirroring process.

xichen2020: I'd think the process would be to first generate the placement, and the agg tier applies the placement, and finally the agg tier cleans up the shard states after the placement is applied.

For a node replace, there'll be a period in which both the old and the new instances are serving the same shard set and therefore dual ingesting data, which temporarily bumps up the RF for that shard set. Not sure I understand why that would be an issue for the mirroring process as the placement would have been already generated at that point.

cw9: that's correct, what I meant is, for example, when the replace just happened, the RF for the shardset became 3 at that point, but the placement is supposed to have a shardset of 2, now you realize the new instance is having problem and could not take shards over/mark them as available. Now you want to revert the replace, so you call the replace function reversely again, if you look at the current algorithm, in the beginning of each operation, it has to generate a mirror placement from the original placement, in this case, it would report an error when trying to generate the mirror for the shardset because it has 3 instances for the shardset while the algorithm is expecting only 2, also the 3 instances have different status in the shards, so I have to clean up the shard states from previous placement changes before applying the next change.

cw9: This is why you could mark shards as available as the topology change rolls out, but before you make the next topology change, the placement must be cleaned up, otherwise the algo will report error, and since we can't guarantee that placement can always be cleaned up automatically, I'm doing that in the beginning of each operation, as long as you leave enough time between topology changes, I don't think there is any harm of doing that.

xichen2020: Understood, but if we back out of a replace before the replacement finishes, won't you still have three replicas for the shardset owned by the replacement node regardless of whether you mark all the shard states as available?

cw9: As you mark all shards as available, the leaving instance will be removed from the placement because it will no longer own any shards after all the marking. After that the shardSetID will have the same number of instances as the rest of the placement.

services/placement/algo/mirrored.go
cw9: I'd prefer to put that logic outside of the algo because it involves instance filtering, for example if you pass in 15 instances to be added to a placement with rf=2, you can at most make 7 pairs of instances and 1 instance will be left out, I prefer to make the algo to be able to take in all instances passed in and return error otherwise. So it's better for the placement service to decide which 14 instances to make the 7 pairs of instances, and while it's doing that, assigning shardSetID to each pair is pretty easy.

services/placement/service/placementservice.go
cw9: sure

services/placement/algo/mirrored.go
xichen2020: Okay fair enough.

services/placement/algo/mirrored.go
xichen2020: Hm say if we back out of an unfinished replace, if we always remove the leaving instance from the placement after marking shards available, will that work when we introduce shard-level cutover and cutoff times?

cw9: Say you replace instance A with instance B, shards on instance A will be in leaving state and on instance B will be in Initializing state. If you back out of this unfinished replace, meaning you must have called the placement algo to replace B with A, the algorithm will first mark all shards as available, so instance A will be removed temporarily and all shards on instance B will be in available state, then replace B with A. 

So in the end of the "back out" you will see all shards on instance A as Initializing and on instance B as Leaving state. the clients will still be able to dual write to both.

xichen2020: Ok yeah seems reasonable.

PR #139, opened 2017-08-09
Support filtering instances for mirror placement

cw9:
Add a InstanceFilter interface, moved most of the existing instance filtering logic from placement service to filter/non_mirrored.go, added filter/mirrored.go to support instance filtering for mirrored placement.

For Initializing placement, the filter will try to make as many pairs as possible from the provided candidate list. 
For adding instances, the filter will stop when first pair of instances found to be added.
For replacing instances, the filter will validate that all leaving instances are from the same host, and then find a valid host from the candidate list to replace all the leaving instances.

@xichen2020


services/placement/filter/filter.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package filter
```
xichen2020: Looking through the usage of the `Filter` interface, it seems the filter is used to algorithmically select the most appropriate instances for various operations, therefore I feel it's more appropriate to call the interface `Selector`, with `SelectAddingInstances` and `SelectReplaceInstances` etc defined on the interface.

services/placement/filter/non_mirrored.go
```diff
@@ -0,0 +1,317 @@
...
+			continue
+		}
+		if placement.IsInstanceLeaving(instanceInPlacement) {
+			instances = append(instances, instanceInPlacement)
+		}
+	}
+
+	instances = filterZones(p, instances, opts)
+	if len(instances) == 0 {
+		return nil, errNoValidNewInstance
```
xichen2020: nit: `errNoValidCandidateInstance` might be better?

services/placement/types.go
```diff
@@ -36,13 +36,39 @@ type Algorithm interface {
...
+		leavingInstanecIDs []string,
+		addingInstances []services.PlacementInstance,
+	) (services.Placement, error)
 
 	// IsCompatibleWith checks whether the algorithm could be applied to given placement.
 	IsCompatibleWith(p services.Placement) error
 }
 
+// InstanceFilter filters valid instances for the placement change.
+type InstanceFilter interface {
```
xichen2020: As mentioned above, would call this `InstanceSelector`, with `Filter` replaced with `Selector` for the methods defined in the interface.

services/placement/algo/sharded.go
```diff
@@ -143,46 +143,53 @@ func (a rackAwarePlacementAlgorithm) AddInstances(
...
 	if err != nil {
 		return nil, err
 	}
 
-	err = ph.PlaceShards(leavingInstance.Shards().All(), leavingInstance, addingInstances)
-	if err != nil && err != errNotEnoughRacks {
-		return nil, err
+	for _, leavingInstance := range leavingInstances {
+		err = ph.PlaceShards(leavingInstance.Shards().All(), leavingInstance, addingInstances)
+		if err != nil && err != errNotEnoughRacks {
```
xichen2020: Right I see, can we put a comment here to clarify?

services/placement/filter/mirrored.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	var pairs = make([][]services.PlacementInstance, 0, len(candidates))
+	for _, hosts := range weightToHostMap {
+		pairedHosts := pairHosts(hosts, p.ReplicaFactor())
+		if len(pairedHosts) == 0 {
+			continue
+		}
+
+		pairs, err = pairInstances(pairedHosts[:1])
```
xichen2020: Right I should've clarified, I meant when we got an error here when grouping instances, do we wanna continue trying other groups too instead of bailing early, given that there may be eligible instances in other groups that could satisfy the request?

Also is there any logic that prefers instances with larger or smaller weights?

services/placement/filter/mirrored.go
```diff
@@ -0,0 +1,435 @@
...
+		ssIDs = make(map[string]struct{}, len(leavingInstances))
+	)
+	for _, instance := range leavingInstances {
+		if h.name == "" {
+			h.name = instance.Hostname()
+			h.weight = instance.Weight()
+			h.rack = instance.Rack()
+			h.portToInstance = make(map[string]services.PlacementInstance, len(leavingInstances))
+		}
+		if h.weight != instance.Weight() {
```
xichen2020: You can still return an error from the `Equals` method and include details in the error, no?

services/placement/filter/mirrored.go
```diff
@@ -0,0 +1,435 @@
...
+	return instances
+}
+
+func nextNShardSetIDs(p services.Placement, n int) []string {
+	var (
+		instances          = p.Instances()
+		currentShardSetIDs = make(map[int]struct{}, len(instances))
+	)
+	for _, instance := range instances {
+		ssID, err := strconv.Atoi(instance.ShardSetID())
```
xichen2020: The `shardSetIDAfter(ssid string)` would simply be passed in from the options and has no default implementation. Actually you could have a `ShardSetIDGenerator` interface with a `NewShardSetIDGenerator(startID string)` constructor and a `NextID() (string, error)` method that simply returns the next id, which might be cleaner. This way the generator can generate uuids if users prefer to.

Don't think it's a big concern for not reusing shard set ids, given that it could be arbitrary strings.

services/placement/filter/mirrored.go
```diff
@@ -0,0 +1,435 @@
...
+	weight         uint32
+	portToInstance map[string]services.PlacementInstance
+}
+
+type rack struct {
+	rack  string
+	hosts []host
+}
+
+type rackHeap struct {
```
xichen2020: The slight inconvenience of dereferencing pointers is probably worth the effort. It saves you some typing too. FWIW this is the example in the `container/heap` package: https://play.golang.org/p/Vz_mpb85uR.

services/placement/algo/sharded.go
cw9: Yeah sure

services/placement/filter/mirrored.go
cw9: hmm, probably not, the only reason this could error out is when hosts have different ports available, which should never happen. Also when that happens we may want to fail early and debug what caused the issue(most likely the r2admin created bad instances).

Right now there is no logic that prefers weights, not sure if we want to prefer larger or smaller weight, but the logic could be added.

services/placement/filter/mirrored.go
cw9: I was comparing weight in another place and used a different error msg, but sure can combine the 2 cases.

services/placement/filter/mirrored.go
cw9: Sounds good, will change

services/placement/filter/mirrored.go
xichen2020: I see sounds good, I'm fine with punting on the weight preference logic since the machines in the same pool are likely to (or should rather) have the same weight.

services/placement/selector/mirrored.go
```diff
@@ -0,0 +1,465 @@
...
+
+import (
+	"container/heap"
+	"errors"
+	"fmt"
+	"math"
+	"strconv"
+
+	"github.com/m3db/m3x/log"
+
```
xichen2020: Remove the blank line?

services/placement/selector/mirrored.go
```diff
@@ -0,0 +1,465 @@
...
+	weightToHostMap, err := groupHostsByWeight(candidates)
+	if err != nil {
+		return nil, err
+	}
+
+	var groups = make([][]services.PlacementInstance, 0, len(candidates))
+	for _, hosts := range weightToHostMap {
+		groupedHosts, ungrouped := groupHostsWithRackCheck(hosts, rf)
+		if len(ungrouped) != 0 {
+			for _, host := range ungrouped {
```
xichen2020: If there are ungrouped hosts for the initial placement, should we return an error instead?

services/placement/selector/mirrored.go
```diff
@@ -0,0 +1,465 @@
...
+	for _, h := range hosts {
+		r, ok := uniqRacks[h.rack]
+		if !ok {
+			r = &rack{
+				rack:  h.rack,
+				hosts: make([]host, 0, rf),
+			}
+
+			uniqRacks[h.rack] = r
+			rh.Push(r)
```
xichen2020: I think you should consistently use `heap.Push(rh, r)` instead of member methods, which is what the `container/heap` package example suggests. Can you do a sweep across the PR and make sure that's the case?

services/placement/selector/mirrored.go
```diff
@@ -0,0 +1,465 @@
...
+		// When there are more than rf racks available, try to make a group.
+		seenRacks := make(map[string]*rack, rf)
+		groups := make([]host, 0, rf)
+		for i := 0; i < rf; i++ {
+			r := heap.Pop(&rh).(*rack)
+			// Move the host from the rack to the group. The racks in the heap
+			// always have at least one host.
+			groups = append(groups, r.hosts[len(r.hosts)-1])
+			r.hosts = r.hosts[:len(r.hosts)-1]
+			seenRacks[r.rack] = r
```
xichen2020: We no longer need to check if we've seen the rack before?

services/placement/selector/mirrored.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+type racksByNumHost []*rack
+
+func (h racksByNumHost) Len() int {
+	return len(h)
+}
+
+func (h racksByNumHost) Less(i, j int) bool {
+	return len(h[i].hosts) > len(h[j].hosts)
```
xichen2020: We now have unit tests to test this yeah?

services/placement/selector/mirrored.go
cw9: Depend on how we plan to use it, if we want to just init from a pool of hosts, say we have some (2x+1) hosts in the pool, then it's very likely that some hosts will be left ungrouped.

But if we always give a list of hosts we probably want to use every host in the initial placement, then we should get an error if some hosts could not be grouped.

services/placement/selector/mirrored.go
cw9: This is before the heap.Init(), I was just appending the rack to the list and don't want to confuse that this needs to be a heap push.

services/placement/selector/mirrored.go
cw9: Nope, the rack is unique in the heap and won't exist in the map

services/placement/selector/mirrored.go
cw9: yes

cw9:
> and pull in the change to make shard set id ints, should be good to go.

Do you want me to just change it to ints? Would it break your current placement usage?

services/placement/selector/mirrored.go
xichen2020: Okay yeah I think mostly likely we'll pull hosts from a pool, so I'm fine w/ leaving it as is.

services/placement/selector/mirrored.go
xichen2020: I see, in that case can you do `rh = append(rh, r)` instead of using a heap member method? More clear that way.

services/placement/selector/mirrored.go
xichen2020: Cool sounds good.

xichen2020:
Put up a separate PR: https://github.com/m3db/m3cluster/pull/144 for this. Would rather not bundle together multiple changes in one PR.

services/placement/types.go
```diff
@@ -36,13 +36,39 @@ type Algorithm interface {
 	// RemoveInstances removes a list of instances from the placement.
 	RemoveInstances(p services.Placement, leavingInstanceIDs []string) (services.Placement, error)
 
-	// ReplaceInstance replace a instance with new instances.
-	ReplaceInstance(p services.Placement, leavingInstanceID string, addingInstances []services.PlacementInstance) (services.Placement, error)
+	// ReplaceInstance replace a list of instances with new instances.
+	ReplaceInstances(
```
prateek: nit:  retain the old method, and have it call the new extended method. will be good to not break all the dependent code (e.g. m3admin, dtests)

services/placement/selector/mirrored.go
cw9: Sure, sounds good

services/placement/types.go
cw9: Yeah... I could do that, I was trying to take the chance to keep less code, if you think the effort of fixing the broken dependent is too much I'm ok to keep the old interface.

PR #141, opened 2017-08-15
Add shard-level cutover/cutoff for staged shard transitions

xichen2020:
cc @cw9 @prateek  @martin-mao 

This PR adds shard-level cutover and cutoff times to support staged shard transitions for agg tier topology changes.

I could've instead used a single `state_change_at` field in the shard proto message for some space savings and infer cutover / cutoff times based on a combination of shard state and state change times but this feels more explicit and requires no custom logic. If needed, the placement generation logic can always zero out cutover/cutoff times to save space that have take effect since protobuf doesn't persist fields with default values.


cw9:
Sounds good, while you are at it, could you change the shard_set_id to uint32 when you are generating the new placement for agg tier? I saw it being used as uint32 in m3agg diff and in my other diff I have to convert the string to uint32 in order to find next shard_set_ids when adding new instances.

xichen2020:
@cw9 The reason why I mad the shard set id strings was to make it slightly more generic so it can be arbitrary strings (e.g., uuid, strings containing some metadata information, etc). Can you take in a id factory function that takes in a seed and generates new ids?

cw9:
I remember you were directly assuming shardSetID as numbers in an m3agg diff? shardSetID as string still works for me, I just felt it can be simpler in many cases if shardSetID is a integer.

xichen2020:
I was using it as string in m3agg PRs and other places. If we feel the need to make the shard set id more generic than ints is unnecessary for other use cases, I can change it back to ints.

cw9:
Let's just keep it as string for now

shard/shard.go
```diff
@@ -80,6 +80,18 @@ type Shard interface {
 	// ID returns the ID of the shard
 	ID() uint32
 
+	// CutoverNanos returns when shard traffic is cut over.
+	CutoverNanos() int64
```
prateek: Feels weird to have the same interface used by `m3db` and `m3aggregator` if these fields are only used by one of them. Should they be separated?

xichen2020: The reason why they live in the `Shard` interface is that the placement algorithm already works with the shard interface, so if we separate them out into different types, we need to either duplicate existing algorithm code to work with a different type, or have type casts in various places in the algorithm to do different things based on the shard type, which is also messy.

Adding these two methods to the shard interface makes the shards slightly more generic and is also non-intrusive to support the shard-level cutover / cutoff we need without significantly rewriting the algorithms. Even For `m3db`, if it needs staged shard cutover or cutoff at some point for whatever reason, it's readily supported.

PR #143, opened 2017-08-19
[WIP][gometalinter] Add gometalinter

jeromefroe:
This PR adds `gometalinter` to the `m3cluster` repo and addresses any issues it finds.

cc @xichen2020 @prateek

### DIFF
--- a/client/etcd/config_test.go
+++ b/client/etcd/config_test.go
@@ -66,6 +66,7 @@ func TestConfig(t *testing.T) {
 	require.Equal(t, 10*time.Second, cfg.SDConfig.InitTimeout)
 }
 
+// nolint: unparam
 func writeFile(t *testing.T, contents string) string {
 	f, err := ioutil.TempFile("", "configtest")
 	require.NoError(t, err)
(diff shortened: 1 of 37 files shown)


client/etcd/config_test.go
```diff
@@ -66,6 +66,7 @@ func TestConfig(t *testing.T) {
 	require.Equal(t, 10*time.Second, cfg.SDConfig.InitTimeout)
 }
 
+// nolint: unparam
```
prateek: Weird that this reports an error.

kv/etcd/store.go
```diff
@@ -391,8 +391,7 @@ func (c *client) update(key string, events []*clientv3.Event) error {
 		err error
 	)
 
-	noEvents := events == nil || len(events) == 0
-	if noEvents {
+	if len(events) == 0 {
```
cw9: interesting, doesn't len(events) also cover the events == nil case?

client/etcd/config_test.go
jeromefroe: Yea, it's because we always pass the same string for `contents`, so it suggests removing it as a parameter.

PR #145, opened 2017-08-23
[util] Add utility functions for validating updates to watch variables

jeromefroe:
This PR adds utility functions to `kv` to watch for updates and validate that the updates fall within the pre-specified bounds. These functions will promote defensive programming and help callers guard against incorrect values being set in KV. I took a narrow approach here, only validating bounds for numeric types, but it might make sense to expand these changes to a more generic `validate` function that can check any type.

Update: Decided to take Prateek's advice and make the validation function more agnostic to the validation that should be performed.

cc @cw9 @xichen2020 @prateek

### DIFF
--- a/kv/util/util.go
+++ b/kv/util/util.go
@@ -22,7 +22,6 @@ package util
 
 import (
 	"errors"
-	"fmt"
 	"sync"
 	"time"
 
@@ -33,11 +32,12 @@ import (
 
 var errNilStore = errors.New("kv store is nil")
 
+// ValidateFn validates an update from KV.
+type ValidateFn func(interface{}) error
+
 type getValueFn func(kv.Value) (interface{}, error)
 
 type updateFn func(interface{})
-
-type validateBoundsFn func(interface{}) (lower, upper interface{}, err error)
 
 // WatchAndUpdateBool sets up a watch for a bool property.
 func WatchAndUpdateBool(
@@ -52,7 +52,7 @@ func WatchAndUpdateBool(
 	return watchAndUpdate(store, key, getBool, updateFn, nil, defaultValue, logger)
 }
 
-// WatchAndUpdateFloat64 sets up a watch for an float64 property.
+// WatchAndUpdateFloat64 sets up a watch for a float64 property.
 func WatchAndUpdateFloat64(
 	store kv.Store,
 	key string,
@@ -61,9 +61,7 @@ func WatchAndUpdateFloat64(
 	defaultValue float64,
 	logger xlog.Logger,
 ) (kv.ValueWatch, error) {
-	updateFn := lockedUpdate(func(i interface{}) { *property = i.(float64) }, lock)
-
-	return watchAndUpdate(store, key, getFloat64, updateFn, nil, defaultValue, logger)
+	return WatchAndUpdateWithValidationFloat64(store, key, property, lock, defaultValue, nil, logger)
 }
 
 // WatchAndUpdateInt64 sets up a watch for an int64 property.
@@ -75,12 +73,10 @@ func WatchAndUpdateInt64(
 	defaultValue int64,
 	logger xlog.Logger,
 ) (kv.ValueWatch, error) {
-	updateFn := lockedUpdate(func(i interface{}) { *property = i.(int64) }, lock)
-
-	return watchAndUpdate(store, key, getInt64, updateFn, nil, defaultValue, logger)
+	return WatchAndUpdateWithValidationInt64(store, key, property, lock, defaultValue, nil, logger)
 }
 
-// WatchAndUpdateString sets up a watch for an string property.
+// WatchAndUpdateString sets up a watch for a string property.
 func WatchAndUpdateString(
 	store kv.Store,
 	key string,
@@ -89,9 +85,7 @@ func WatchAndUpdateString(
 	defaultValue string,
 	logger xlog.Logger,
 ) (kv.ValueWatch, error) {
-	updateFn := lockedUpdate(func(i interface{}) { *property = i.(string) }, lock)
-
-	return watchAndUpdate(store, key, getString, updateFn, nil, defaultValue, logger)
+	return WatchAndUpdateWithValidationString(store, key, property, lock, defaultValue, nil, logger)
 }
 
 // WatchAndUpdateTime sets up a watch for a time property.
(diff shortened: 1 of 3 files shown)


kv/util/util.go
```diff
@@ -33,54 +34,179 @@ import (
 var errNilStore = errors.New("kv store is nil")
 
 type getValueFn func(kv.Value) (interface{}, error)
+
 type updateFn func(interface{})
 
+type validateBoundsFn func(interface{}) (lower, upper interface{}, err error)
```
prateek: More generic to have a `validateFn` with signature: 

```
type validateFn func(interface{}) error 
```

It'll be type agnostic and allow for more custom validation if we ever need it

jeromefroe: Yea, I'll update this. I went back and forth myself. Think a more agnostic `validate` is the way to go.

kv/util/util.go
```diff
@@ -256,6 +407,25 @@ func logInvalidUpdate(logger xlog.Logger, k string, ver int, v interface{}, err
...
+	err error,
+) {
+	getLogger(logger).WithFields(
+		xlog.NewLogField("key", k),
+		xlog.NewLogField("value", v),
+		xlog.NewLogField("version", ver),
+		xlog.NewLogField("lower-bound", lower),
+		xlog.NewLogField("upper-bound", upper),
+		xlog.NewLogField("error", err),
+	).Infof(
```
prateek: nit: `Warnf` instead of `Infof`

kv/util/util.go
```diff
@@ -247,13 +360,22 @@ func logUpdateSuccess(logger xlog.Logger, k string, ver int, v interface{}) {
...
+}
+
 func logInvalidUpdate(logger xlog.Logger, k string, ver int, v interface{}, err error) {
 	getLogger(logger).WithFields(
 		xlog.NewLogField("key", k),
 		xlog.NewLogField("value", v),
 		xlog.NewLogField("version", ver),
 		xlog.NewLogField("error", err),
-	).Infof("invalid value from kv store, use default value")
+	).Warnf("invalid value from kv store, not applying update")
```
xichen2020: The second part of the warning message should probably be consistent with `logMalformedUpdate`?

kv/util/util.go
```diff
@@ -32,55 +32,144 @@ import (
...
+) (kv.ValueWatch, error) {
+	return WatchAndUpdateWithValidationTime(store, key, property, lock, defaultValue, nil, logger)
 }
 
-// WatchAndUpdateFloat64 sets up a watch for an float64 property.
-func WatchAndUpdateFloat64(store kv.Store, key string, property *float64, lock sync.Locker,
-	defaultValue float64, logger xlog.Logger) (kv.ValueWatch, error) {
+// WatchAndUpdateWithValidationFloat64 sets up a watch with validation for a float64
+// property. Any invalid updates are not applied.
+func WatchAndUpdateWithValidationFloat64(
```
xichen2020: Can you do a search across OSS and internal repos to see how many places you'd need to change if users are always required to pass in validation function (`nil` if no validation function)? Having both versions maintains API backward compatibility at the cost of wider and more verbose public interface so I'd prefer to just merge them if it doesn't take too much effort.

jeromefroe: Yea, meant to follow up on this after standup, I think that's the better approach, will take a look across OSS and statsdex.

kv/util/util.go
jeromefroe: We should discuss this I think. I don't feel that we should be resetting the value to the default value for malformed updates. That seems wrong to me since if the update is bad there really isn't any reason to apply the update at all in my opinion. And I feel even more strongly that we should not apply the default value for invalid updates. So if we are going to make the behavior consistent I'm in favor of changing the logic for malformed updates to not apply the default value.

kv/util/util.go
```diff
@@ -223,10 +282,18 @@ func updateWithKV(getValue getValueFn, update updateFn, key string, v kv.Value,
 	newValue, err := getValue(v)
 	if err != nil {
 		update(defaultValue)
-		logInvalidUpdate(logger, key, v.Version(), defaultValue, err)
+		logMalformedUpdate(logger, key, v.Version(), defaultValue, err)
 		return err
 	}
 
+	if validate != nil {
```
cw9: Feels a little weird that a malformed update applies the default value while a well-formed invalid value does not, maybe we should change the behavior to not apply default value in both cases?

jeromefroe: Sounds good to me, will update. I had left it this way since we applied the defaults for malformed previously, so I didn't want to change it until we talked about it.

kv/util/util.go
```diff
@@ -238,57 +321,65 @@ func watchAndUpdate(
...
-	defaultValue interface{},
+	validate ValidateFn,
 	logger xlog.Logger,
 ) error {
 	if v == nil {
-		// The key is deleted from kv, use the default value.
-		update(defaultValue)
-		logSetDefault(logger, key, defaultValue)
-		return nil
+		logNilUpdate(logger, key)
```
cw9: I think we should update default value in this case? this(deletion) should actually be a valid case?

kv/util/util.go
```diff
@@ -30,144 +31,226 @@ import (
...
+)
 
-// WatchAndUpdateBool sets up a watch with validation for a bool property. Any invalid
-// updates are not applied.
+// WatchAndUpdateBool sets up a watch with validation for a bool property. Any
+// empty, malformed, or invalid updates are not applied.
 func WatchAndUpdateBool(
 	store kv.Store,
 	key string, property *bool,
 	lock sync.Locker,
```
cw9: nit: lock can be moved to Options as well? the code supports cases with no lock already

kv/util/util.go
```diff
@@ -238,57 +321,65 @@ func watchAndUpdate(
...
-		return nil, err
+		return fmt.Errorf("could not establish initial watch: %v", err)
 	}
 
 	go func() {
 		for {
-			_, ok := <-watch.C()
-			if !ok {
-				return
+			if watch == nil {
```
cw9: don't think you'll ever hit this case, the contract of the kv.ValueWatch is to fix itself once you get it. So users don't need to create it.

You could potentially add a error log in the !ok case, although that case should not never happen as well.

jeromefroe: Unless you feel strongly, I would actually prefer to leave it as is, and a comment that this case should never happen. My concern is that if we ever change the contract of `kv.ValueWatch` we'll change the behavior here as well unless we're careful.

kv/util/util.go
jeromefroe: I thought about this originally, but I couldn't think of any case where you wouldn't want to pass a lock. Since we update the lock in a separate goroutine, without a lock you're introducing a race detection. If we wanted to support a lockless case I would prefer adding functions which performs their writes atomically in a separate PR.

kv/util/util.go
jeromefroe: Yea, I wasn't sure about this. I feel that we can guard against accidentally deleting a key in kv by not applying an update, but then I can also imagine cases where you will want to watch a key but it need not exist. It seems that currently we only watch keys that we expect to exist, so maybe we should include in the comments that the key should always exist so we don't have to address the case of optional keys here and can instead address when the need arises with a different set of functions for that use case?

kv/util/util.go
cw9: I'd prefer to rely on the Watch on doing this since that's the contract, the watch itself is doing the recreation more comprehensively, if the contract is broken then we have plenty places that will be broken as well. I think an error log is more than enough here.

kv/util/util.go
cw9: Not really, there are plenty keys that don't exist, for example when we create a new environment, we usually just copy some keys with custom values to the new environment, otherwise they rely on this case to get the default value.

kv/util/util.go
cw9: Sure, sounds good.

kv/util/util.go
jeromefroe: That's a good point and it would simplify the code here. Will update to add a comment on the contract of `kv.ValueWatch` and log an error.

kv/util/util.go
jeromefroe: Ahh, ok, if it's customary to use the watches with keys that don't exist, then I'll update. Wasn't aware that was the case.

kv/util/util.go
```diff
@@ -390,10 +410,11 @@ func updateWithKV(
 	return nil
 }
 
-func logNilUpdate(logger xlog.Logger, k string) {
+func logNilUpdate(logger xlog.Logger, k string, v interface{}) {
 	getLogger(logger).WithFields(
 		xlog.NewLogField("key", k),
-	).Warnf("nil value from kv store, not applying update")
+		xlog.NewLogField("value", v),
```
cw9: nit: "value" to "default value"

cw9: lol, seems you updated in the next commit, nvm

PR #146, opened 2017-08-25
Support custom namespace override in service discovery

cw9:
Right now by default the placement data and metadata data in service discovery service are stored under _sd.placement and _sd.metadata namespace. This pr adds a new method to allow users to provide custom namespace to override those namespaces. So that we could support the r2 placement being accessible from both ServiceDiscovery interface and kv.Store interface although the value is stored in a custom namespace on etcd.

For example if a namespace of "/test_placement" was provided, the placement data for service: "statsdex_m3dbnode" could be accessible under "/test_placement/[env]/statsdex_m3dbnode".

In the future we could allow overriding the default namespace for leader service and heartbeat service as well.

@xichen2020 @prateek @jeromefroe @schallert @dgromov

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -29,10 +29,7 @@ import (
 // access to cluster services.
 type Client interface {
 	// Services returns access to the set of services.
-	Services() (services.Services, error)
-
-	// ServiceDiscovery returns service discovery with custom namespaces.
-	ServiceDiscovery(opts services.Options) (services.Services, error)
+	Services(opts services.Options) (services.Services, error)
 
 	// KV returns access to the distributed configuration store.
 	// To be deprecated.
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -91,9 +91,9 @@ type csclient struct {
 	logger  xlog.Logger
 	newFn   newClientFn
 
-	sdOnce sync.Once
-	sd     services.Services
-	sdErr  error
+	defaultServicesOnce sync.Once
+	defaultServices     services.Services
+	defaultServicesErr  error
 
 	txnOnce sync.Once
 	txn     kv.TxnStore
@@ -101,11 +101,11 @@ type csclient struct {
 }
 
 func (c *csclient) Services() (services.Services, error) {
-	c.sdOnce.Do(func() {
-		c.sd, c.sdErr = c.createServices(services.NewOptions())
+	c.defaultServicesOnce.Do(func() {
+		c.defaultServices, c.defaultServicesErr = c.createServices(services.NewOptions())
 	})
 
-	return c.sd, c.sdErr
+	return c.defaultServices, c.defaultServicesErr
 }
 
 func (c *csclient) ServiceDiscovery(opts services.Options) (services.Services, error) {
@@ -144,7 +144,7 @@ func (c *csclient) createServices(opts services.Options) (services.Services, err
 		SetHeartbeatGen(c.heartbeatGen()).
 		SetKVGen(c.kvGen(c.cacheFileFn(cacheFileExtraFields...))).
 		SetLeaderGen(c.leaderGen()).
-		SetNamespaceOptions(opts.NamespaceOptions()).
+		SetNamespaceOptions(nOpts).
 		SetInstrumentsOptions(instrument.NewOptions().
 			SetLogger(c.logger).
 			SetMetricsScope(c.sdScope),
@@ -255,14 +255,16 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
 
 func (c *csclient) cacheFileFn(extraFields ...string) cacheFileForZoneFn {
 	return func(zone string) etcdkv.CacheFileFn {
-		return etcdkv.CacheFileFn(func(namespace string) string {
+		return func(namespace string) string {
 			if c.opts.CacheDir() == "" {
 				return ""
 			}
-			cacheFileFields := []string{namespace, c.opts.Service(), zone}
+
+			cacheFileFields := make([]string, 0, len(extraFields)+3)
+			cacheFileFields = append(cacheFileFields, namespace, c.opts.Service(), zone)
 			cacheFileFields = append(cacheFileFields, extraFields...)
 			return filepath.Join(c.opts.CacheDir(), fileName(cacheFileFields...))
-		})
+		}
 	}
 }
 
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -100,15 +100,7 @@ type csclient struct {
 	txnErr  error
 }
 
-func (c *csclient) Services() (services.Services, error) {
-	c.defaultServicesOnce.Do(func() {
-		c.defaultServices, c.defaultServicesErr = c.createServices(services.NewOptions())
-	})
-
-	return c.defaultServices, c.defaultServicesErr
-}
-
-func (c *csclient) ServiceDiscovery(opts services.Options) (services.Services, error) {
+func (c *csclient) Services(opts services.Options) (services.Services, error) {
 	return c.createServices(opts)
 }
 
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -55,6 +55,8 @@ var errInvalidNamespace = errors.New("invalid namespace")
 
 type newClientFn func(endpoints []string) (*clientv3.Client, error)
 
+type cacheFileForZoneFn func(zone string) etcdkv.CacheFileFn
+
 // NewConfigServiceClient returns a ConfigServiceClient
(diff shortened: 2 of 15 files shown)


xichen2020:
Can you add a brief description w.r.t. what this PR does and what the main changes are? This way the reviewer can quickly understand its purpose and pay more attention to the main changes.

cw9:
Updated pr description

client/etcd/client_test.go
```diff
@@ -23,10 +23,10 @@ package etcd
 import (
 	"testing"
 
-	"github.com/m3db/m3cluster/services"
-
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/integration"
+	"github.com/m3db/m3cluster/services"
```
xichen2020: This should be above 3rd party imports with blank line in between.

client/client.go
```diff
@@ -31,6 +31,9 @@ type Client interface {
 	// Services returns access to the set of services.
 	Services() (services.Services, error)
 
+	// ServiceDiscovery returns service discovery with custom namespaces.
+	ServiceDiscovery(opts services.Options) (services.Services, error)
```
xichen2020: Any reason why we have a `ServiceDiscovery` API in addition to the `Services` API? Would it be possible to merge the two into `Services(opts services.Options)` and if the options are the default options and there's no namespace override, we return the default services?

client/etcd/client.go
```diff
@@ -123,40 +137,37 @@ func (c *csclient) TxnStore(namespace string) (kv.TxnStore, error) {
...
-		)
-	})
+func (c *csclient) createServices(opts services.Options) (services.Services, error) {
+	nOpts := opts.NamespaceOptions()
+	cacheFileExtraFields := []string{nOpts.PlacementNamespace(), nOpts.MetadataNamespace()}
+	return sdclient.NewServices(c.sdOpts.
+		SetHeartbeatGen(c.heartbeatGen()).
+		SetKVGen(c.kvGen(c.cacheFileFn(cacheFileExtraFields...))).
+		SetLeaderGen(c.leaderGen()).
+		SetNamespaceOptions(opts.NamespaceOptions()).
```
xichen2020: nit: just `nOpts`

client/etcd/client.go
```diff
@@ -242,13 +253,17 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
...
-		}
-		return filepath.Join(c.opts.CacheDir(), fileName(namespace, c.opts.Service(), zone))
-	})
+func (c *csclient) cacheFileFn(extraFields ...string) cacheFileForZoneFn {
+	return func(zone string) etcdkv.CacheFileFn {
+		return etcdkv.CacheFileFn(func(namespace string) string {
+			if c.opts.CacheDir() == "" {
+				return ""
+			}
+			cacheFileFields := []string{namespace, c.opts.Service(), zone}
```
xichen2020: nit: preallocate enough space for the slice.

client/etcd/client.go
```diff
@@ -242,13 +253,17 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
...
 }
 
-func (c *csclient) cacheFileFn(zone string) etcdkv.CacheFileFn {
-	return etcdkv.CacheFileFn(func(namespace string) string {
-		if c.opts.CacheDir() == "" {
-			return ""
-		}
-		return filepath.Join(c.opts.CacheDir(), fileName(namespace, c.opts.Service(), zone))
-	})
+func (c *csclient) cacheFileFn(extraFields ...string) cacheFileForZoneFn {
```
xichen2020: Are `extraFields` essentially namespaces?

client/etcd/client.go
xichen2020: Also, what's the `namespace` in this case? In particular, what would a cache file path look like for the cache path containing m3aggregator placement key, assuming the placement namespace is `/r2/placement`?

client/etcd/client.go
```diff
@@ -97,11 +101,17 @@ type csclient struct {
 }
 
 func (c *csclient) Services() (services.Services, error) {
-	c.createServices()
+	c.sdOnce.Do(func() {
+		c.sd, c.sdErr = c.createServices(services.NewOptions())
```
xichen2020: Perhaps rename `Services` to `DefaultServices` and `sd` and `sdErr` to `defaultServices` and `defaultServicesErr`?

client/etcd/client.go
```diff
@@ -242,13 +253,17 @@ func newClient(endpoints []string) (*clientv3.Client, error) {
...
-func (c *csclient) cacheFileFn(zone string) etcdkv.CacheFileFn {
-	return etcdkv.CacheFileFn(func(namespace string) string {
-		if c.opts.CacheDir() == "" {
-			return ""
-		}
-		return filepath.Join(c.opts.CacheDir(), fileName(namespace, c.opts.Service(), zone))
-	})
+func (c *csclient) cacheFileFn(extraFields ...string) cacheFileForZoneFn {
+	return func(zone string) etcdkv.CacheFileFn {
+		return etcdkv.CacheFileFn(func(namespace string) string {
```
xichen2020: nit: probably no need to cast.

services/config.go
```diff
@@ -0,0 +1,45 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package services
+
+// Configuration configs the Options.
+type Configuration struct {
+	NamespaceOverride NamespaceConfiguration `yaml:"namespaceOverride"`
```
xichen2020: nit: `NamespaceOverrides` might be better? Or maybe just `Namespaces`.

services/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+// NewOptions creates a new Options.
+func (cfg Configuration) NewOptions() Options {
+	return NewOptions().
+		SetNamespaceOptions(cfg.NamespaceOverride.NewOptions())
+}
+
+// NamespaceConfiguration configs the NamespaceOptions.
+type NamespaceConfiguration struct {
+	PlacementNamespace string `yaml:"placementNamespace"`
```
xichen2020: nit: might be better to call this `NamespacesConfiguration`, and just call the fields `Placement` and `Metadata`.

services/config_test.go
```diff
@@ -0,0 +1,58 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package services
+
+import (
+	"testing"
+
+	yaml "gopkg.in/yaml.v2"
+
```
xichen2020: Remove blank line here.

client/etcd/client.go
cw9: the cache path will be "/cacheDir/env_service_zone__r2_m3agg.json"

client/client.go
xichen2020: Yeah I'd prefer to merge them to keep the interface a bit cleaner, this isn't too difficult I imagine?

client/etcd/client.go
xichen2020: Did you mean `/cacheDir/env_service_zone__r2_placement.json` if the namespace is `/r2/placement`? Is that still needed for the custom namespace to work the same way as the internal placement namespace?

client/client.go
cw9: yeah, shouldn't be too difficult, gonna merge

client/etcd/client.go
cw9: yes if the namespace is /r2/placement the cache file will be /cacheDir/env_service_zone__r2_placement.json. Yes this is needed so that if you create 2 services client(1 default, 1 with custom placement namespace) their cache file won't override each other.

client/etcd/client.go
```diff
@@ -81,25 +84,24 @@ type csclient struct {
...
 	txnOnce sync.Once
 	txn     kv.TxnStore
 	txnErr  error
 }
 
-func (c *csclient) Services() (services.Services, error) {
-	c.createServices()
-
-	return c.sd, c.sdErr
+func (c *csclient) Services(opts services.Options) (services.Services, error) {
```
xichen2020: Do you want to do sth like
```
if opts == nil {
  opts = services.NewOptions()
}
...
```
so users of default services can just pass in a nil?

PR #148, opened 2017-08-28
Move some code around

cw9:
This pr moves some code around, there is no new code.

1: Rename services/client to services/etcd to keep the same convention as kv/etcd and heartbeat/etcd.

2: Moved placement.NewOptions() to services.NewPlacementOptions() since the interface has been moved to services/

@xichen2020 @prateek @jeromefroe @dgromov @schallert


services/etcd/config.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package client
+package etcd
```
xichen2020: Should the package be `services/client/etcd` instead? This way if in the future there's a non-etcd implementation, it'll go under `services/client/` as well?

services/placement/options.go
```diff
@@ -20,16 +20,8 @@
...
-)
-
 const (
 	defaultMaxStepSize = 3
-	defaultIsSharded   = true
-	// by default partial replace should be allowed for better distribution
-	defaultAllowPartialReplace = true
 )
 
 type deploymentOptions struct {
```
xichen2020: Should the `deploymentOptions` live somewhere else e.g. `services/deploy/`? Does it have to live under `services/placement`?

services/services.go
```diff
@@ -407,3 +411,95 @@ func (opts namespaceOpts) SetMetadataNamespace(v string) NamespaceOptions {
 	opts.metadata = v
 	return opts
 }
+
+type placementOptions struct {
```
xichen2020: Why do we need to move the options here? Feels to me it's better to keep the implementation (and perhaps even the interface) under `services/placement`?

services/etcd/config.go
cw9: sure can do

services/placement/options.go
cw9: DeploymentOptions interface is defined in services/placement so I'd rather leave it here.

services/services.go
cw9: PlacementOptions is already exposed in services/types.go so I'd rather keep the options impl in the services/ package since it's tightly tied to the interface.

xichen2020: I'm not sure if that's the right approach. Feels to me the options implementation is tightly coupled with the placement itself it makes sense to leave them inside `placement`. Why not move the placement options interface to `services/placement`? After all they are placement options so it actually makes sense for them to live inside the `placement` package no?

PR #150, opened 2017-08-30
Set Cutover and Cutoff times for placement and shard in algo

cw9:
This pr adds methods to PlacementOptions interface to allow algorithm to set Cutover and Cutoff times for placement and shard. 


services/placement/algo/sharded_helper.go
```diff
@@ -269,7 +266,9 @@ func (ph *placementHelper) moveShard(candidateShard shard.Shard, from, to servic
 			from.Shards().Remove(shardID)
 			newShard.SetSourceID(candidateShard.SourceID())
 		case shard.Available:
-			candidateShard.SetState(shard.Leaving)
+			candidateShard.
+				SetState(shard.Leaving).
+				SetCutoffNanos(ph.opts.ShardCutoffNanosFn()())
```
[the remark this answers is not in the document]
cw9: Nope, if the newShard is in Unknown state, its CutoverNanos will be set in GeneratePlacement(), if the newShard is Available (due to previously being Leaving) it needs no CutoverNanos.

services/placement/algo/sharded_helper.go
```diff
@@ -723,7 +715,7 @@ func nonLeavingInstances(instances []services.PlacementInstance) []services.Plac
 func newShards(shardIDs []uint32) []shard.Shard {
 	r := make([]shard.Shard, len(shardIDs))
 	for i, id := range shardIDs {
-		r[i] = shard.NewShard(id)
+		r[i] = shard.NewShard(id).SetState(shard.Unknown)
```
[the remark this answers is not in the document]
cw9: Not super necessary, I'm just being safe here in case the NewShard impl changed in the future.

services/placement/options.go
```diff
@@ -50,6 +50,8 @@ func (o deploymentOptions) SetMaxStepSize(stepSize int) DeploymentOptions {
 	return o
 }
 
+func defaultTimeNanosFn() int64 { return 0 }
```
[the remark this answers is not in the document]
cw9: I was gonna handle the case in shard.SetCutoffNanos() because when you get a 0 cutoffNanos from proto, and set that value on a shard, the shard needs to know it means maxInt64 as well, so I'm just going to use this same defaultTimeNanosFn.

This means you can never set a CutoffNanos with 0, but I think that is ok?

services/placement/placement.go
```diff
@@ -538,7 +544,7 @@ func markShardAvailable(p services.Placement, instanceID string, shardID uint32)
 	}
 
 	sourceID := s.SourceID()
-	s.SetState(shard.Available).SetSourceID("")
+	shards.Add(shard.NewShard(shardID).SetState(shard.Available))
```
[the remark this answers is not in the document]
cw9: Add a new shard with same ID will override the original shard with that ID, so here I'm creating a NewShard so SourceID, cutover and cutoff times are all cleaned up.

shard/shard.go
```diff
@@ -33,11 +33,11 @@ import (
 const (
 	// DefaultShardCutoverNanos is the default shard-level cutover nanos
 	// if the value is not set in the proto message.
-	DefaultShardCutoverNanos = 0
+	DefaultShardCutoverNanos = int64(0)
```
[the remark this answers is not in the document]
cw9: I was checking against this value this in a test and got complains that this is a int type not int64

shard/shard.go
```diff
@@ -33,11 +33,11 @@ import (
...
 
 	// DefaultShardCutoffNanos is the default shard-level cutoff nanos
 	// if the value is not set in the proto message.
-	DefaultShardCutoffNanos = math.MaxInt64
+	DefaultShardCutoffNanos = int64(math.MaxInt64)
```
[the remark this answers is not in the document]
cw9: Same

shard/shard.go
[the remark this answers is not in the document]
xichen2020: Right you need to cast in tests , because consts have a default type tests use to compare values, but not in code.

shard/shard.go
```diff
@@ -311,17 +245,40 @@ func (s shards) ShardsForState(state State) []Shard {
...
+
+func (su SortableShardProtosByIDAsc) Len() int           { return len(su) }
+func (su SortableShardProtosByIDAsc) Less(i, j int) bool { return su[i].Id < su[j].Id }
+func (su SortableShardProtosByIDAsc) Swap(i, j int)      { su[i], su[j] = su[j], su[i] }
 
-func (su ShardsByIDAscending) Len() int           { return len(su) }
-func (su ShardsByIDAscending) Less(i, j int) bool { return su[i].Id < su[j].Id }
-func (su ShardsByIDAscending) Swap(i, j int)      { su[i], su[j] = su[j], su[i] }
+// states returns all the possible states.
+func states() []State {
```
xichen2020: nit: `validStates` might be more clear?

shard/types.go
```diff
@@ -0,0 +1,123 @@
...
+	schema "github.com/m3db/m3cluster/generated/proto/placement"
+)
+
+const (
+	// UnInitializedValue is the unintialized value used in proto.
+	UnInitializedValue int64 = 0
+
+	// DefaultShardCutoverNanos is the default shard-level cutover nanos
+	// if the value is not set in the proto message.
+	DefaultShardCutoverNanos int64 = 0
```
xichen2020: You probably don't need to export `DefaultShardCutoverNanos` and `DefaultShardCutoffNanos`?

shard/types.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"math"
+
+	schema "github.com/m3db/m3cluster/generated/proto/placement"
+)
+
+const (
+	// UnInitializedValue is the unintialized value used in proto.
+	UnInitializedValue int64 = 0
```
xichen2020: Use this in the default shard cutover and cutover functions for placement options?

shard/shard.go
```diff
@@ -166,12 +117,28 @@ func (s *shard) CutoffNanos() int64 {
 
 func (s *shard) SetCutoffNanos(value int64) Shard {
 	if value == DefaultShardCutoffNanos {
```
xichen2020: Add comments explaining the rationale behind this?

shard/shard.go
cw9: sure

shard/types.go
cw9: sure

shard/types.go
cw9: I need to check against this in some tests

shard/shard.go
cw9: Sure

PR #152, opened 2017-08-31
More refactoring

cw9:
Moved placement related interfaces from services package to a new placement package.

Removed ServiceID from placement.Storage methods to avoid import cycle between services and placement package. Right now the Storage takes in a ServiceID and can only access placement for this ServiceID.

Renamed proto of placement and metadata to package "placementpb" and "metadatapb" to avoid some issue when generating mocks for placement/types.go, also that's the convention used for some other protos (changesetpb, commonpb, etc).


placement/placement.go
```diff
@@ -533,8 +550,24 @@ func IsInstanceLeaving(instance services.PlacementInstance) bool {
...
-// ByIDAscending sorts PlacementInstance by ID ascending
-type ByIDAscending []services.PlacementInstance
+// Instances is a slice of instances that can produce a debug string.
+type Instances []Instance
+
+func (i Instances) String() string {
+	if len(i) == 0 {
+		return "[]"
+	}
+	var strs []string
```
xichen2020: nit: preallocate enough capacity? Also might wanna use a `bytes.Buffer`?

cw9: bytes.Buffer seems an overkill since this is only used in debugging

PR #153, opened 2017-08-31
Embed placement.Storage in placement.Service

cw9:
placement.Service used to expose Delete(), SetPlacement and Placement(), which are also exposed from placement.Storage interface, now that we also need to expose SetProto() and Proto(), this pr made placement.Storage embedded in placement.Service interface so that future updates in the placement.Storage interface are available through placement.Service as well.

this pr also moved Dryrun logic to placement.Storage.

### DIFF
--- a/placement/service/service.go
+++ b/placement/service/service.go
@@ -32,6 +32,7 @@ import (
 
 type placementService struct {
 	placement.Storage
+
 	opts     placement.Options
 	algo     placement.Algorithm
 	selector placement.InstanceSelector
@@ -88,11 +89,11 @@ func (ps *placementService) BuildInitialPlacement(
 		return nil, err
 	}
 
-	return p, ps.Storage.SetIfNotExist(p)
+	return p, ps.SetIfNotExist(p)
 }
 
 func (ps *placementService) AddReplica() (placement.Placement, error) {
-	p, v, err := ps.Storage.Placement()
+	p, v, err := ps.Placement()
 	if err != nil {
 		return nil, err
 	}
@@ -105,13 +106,13 @@ func (ps *placementService) AddReplica() (placement.Placement, error) {
 		return nil, err
 	}
 
-	return p, ps.Storage.CheckAndSet(p, v)
+	return p, ps.CheckAndSet(p, v)
 }
 
 func (ps *placementService) AddInstances(
 	candidates []placement.Instance,
 ) (placement.Placement, []placement.Instance, error) {
-	p, v, err := ps.Storage.Placement()
+	p, v, err := ps.Placement()
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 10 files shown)


placement/service/service.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 type placementService struct {
-	s        placement.Storage
+	placement.Storage
 	opts     placement.Options
```
xichen2020: nit: blank line after embedded type

placement/service/service.go
```diff
@@ -110,18 +105,13 @@ func (ps placementService) AddReplica() (placement.Placement, error) {
 		return nil, err
 	}
 
-	if ps.opts.Dryrun() {
-		ps.logger.Info("this is a dryrun, the operation is not persisted")
-		return p, err
-	}
-
-	return p, ps.s.CheckAndSet(p, v)
+	return p, ps.Storage.CheckAndSet(p, v)
```
xichen2020: nit: just `ps.CheckAndSet(p, v)`

placement/service/service.go
```diff
@@ -147,16 +137,11 @@ func (ps placementService) AddInstances(
 		addingInstances[i] = addingInstance
 	}
 
-	if ps.opts.Dryrun() {
-		ps.logger.Info("this is a dryrun, the operation is not persisted")
-		return p, addingInstances, err
-	}
-
-	return p, addingInstances, ps.s.CheckAndSet(p, v)
+	return p, addingInstances, ps.Storage.CheckAndSet(p, v)
```
xichen2020: Same, you don't need to explicitly reference `Storage` to call its methods unless the method is overriden in the placement service.

placement/service/service.go
```diff
@@ -110,18 +105,13 @@ func (ps placementService) AddReplica() (placement.Placement, error) {
...
-	return p, ps.s.CheckAndSet(p, v)
+	return p, ps.Storage.CheckAndSet(p, v)
 }
 
-func (ps placementService) AddInstances(
+func (ps *placementService) AddInstances(
 	candidates []placement.Instance,
 ) (placement.Placement, []placement.Instance, error) {
-	p, v, err := ps.s.Placement()
+	p, v, err := ps.Storage.Placement()
```
xichen2020: Same

placement/service/service.go
```diff
@@ -258,31 +233,5 @@ func (ps placementService) MarkInstanceAvailable(instanceID string) error {
 		return err
 	}
 
-	return ps.s.CheckAndSet(p, v)
-}
-
-func (ps placementService) Placement() (placement.Placement, int, error) {
```
xichen2020: We no longer need these in the service interface?

placement/service/service.go
cw9: sg

placement/service/service.go
cw9: sg

placement/service/service.go
cw9: sg

placement/service/service.go
cw9: sg

placement/service/service.go
cw9: We don't, as it's exposed through the embedded Storage interface

PR #154, opened 2017-09-07
Mirror placement algo changes

cw9:
1, Allow replacing instance with Leaving instance in current placement.
2, When removing multiple instances, make sure their shards were only moved to non-leaving instances in the placement.
3. Add Equals() to the Shard interface.

### DIFF
--- a/placement/algo/mirrored.go
+++ b/placement/algo/mirrored.go
@@ -146,15 +146,8 @@ func (a mirroredAlgorithm) AddInstances(
 	}
 
 	// At this point, all leaving instances in the placement are cleaned up.
-	for i, instance := range addingInstances {
-		if _, exist := p.Instance(instance.ID()); exist {
-			return nil, fmt.Errorf("instance %s already exist in the placement", instance.ID())
-		}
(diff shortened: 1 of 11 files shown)


placement/algo/mirrored.go
```diff
@@ -192,6 +197,18 @@ func (a mirroredAlgorithm) ReplaceInstances(
 		return nil, err
 	}
 
+	// At this point, all leaving instances in the placement are cleaned up.
+	for i, instance := range addingInstances {
```
xichen2020: This looks exactly the same as the previous block, refactor it into a common helper function?

shard/shard.go
```diff
@@ -137,6 +137,30 @@ func (s *shard) SetCutoffNanos(value int64) Shard {
 	return s
 }
 
+func (s *shard) Equals(other Shard) bool {
+	if s.ID() != other.ID() {
```
xichen2020: nit: might be cleaner to write
```
return s.ID() == other.ID() &&
  s.State() == other.State() &&
  ...
```

PR #156, opened 2017-09-10
[travis] Add go 1.8 and 1.9 to travis tests

jeromefroe:
This PR adds Go 1.8 and 1.9 to the versions of Go we test against in Travis. 

In addition, it seems the `unparam` linter has been updated and caught some new cases, so this PR also addresses the two instances of an error not being used that it found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.7
-  - 1.8
+  - 1.7.6
+  - 1.8.3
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: go
 go:
   - 1.7
+  - 1.8
```
prateek: same as other PR, change to latest point releases

PR #157, opened 2017-09-11
Support reverting previous change for mirror mode

cw9:
Supports backing out of add/remove/replace for mirror mode.

### DIFF
--- a/placement/algo/mirrored.go
+++ b/placement/algo/mirrored.go
@@ -159,11 +159,10 @@ func (a mirroredAlgorithm) AddInstances(
 	// We just need to get their shards back.
 	if allLeaving(addingInstances, p) {
 		for _, instance := range addingInstances {
-			instance, ok := p.Instance(instance.ID())
-			if !ok {
-				return nil, fmt.Errorf("instance %s not found in placement", instance.ID())
+			ph, instance, err := newAddInstanceHelper(p, instance, a.opts)
+			if err != nil {
+				return nil, err
 			}
-			ph := newAddInstanceHelper(p, instance, a.opts)
 			ph.CleanUpLeavingShards(instance)
 			p = ph.GeneratePlacement()
 		}
@@ -195,8 +194,16 @@ func (a mirroredAlgorithm) AddInstances(
 		if _, ok := mirrorPlacement.Instance(instance.ID()); ok {
 			return nil, fmt.Errorf("shard set id %d already exist in current placement", instance.ShardSetID())
 		}
-		ph := newAddInstanceHelper(mirrorPlacement, instance, a.opts)
-		ph.AddInstance(instance)
+
+		ph, instance, err := newAddInstanceHelper(mirrorPlacement, instance, a.opts)
+		if err != nil {
+			return nil, err
+		}
+
+		if err = ph.AddInstance(instance); err != nil {
+			return nil, err
+		}
+
 		mirrorPlacement = ph.GeneratePlacement()
 	}
 
@@ -212,7 +219,6 @@ func (a mirroredAlgorithm) ReplaceInstances(
 		return nil, err
 	}
 
-	// if allLeaving(addingInstances, p) && allInitializing(instanceIDs, p){
 	if len(addingInstances) != len(leavingInstanceIDs) {
 		return nil, fmt.Errorf("could not replace %d instances with %d instances for mirrored replace", len(leavingInstanceIDs), len(addingInstances))
 	}
--- a/placement/algo/mirrored.go
+++ b/placement/algo/mirrored.go
@@ -99,7 +99,7 @@ func (a mirroredAlgorithm) RemoveInstances(
 			if err != nil {
 				return nil, err
 			}
-			ph.CleanUpInitializingShards(instance)
+			ph.ReturnInitializingShards(instance)
 			p = ph.GeneratePlacement()
 		}
 
@@ -163,7 +163,7 @@ func (a mirroredAlgorithm) AddInstances(
 			if err != nil {
 				return nil, err
 			}
-			ph.CleanUpLeavingShards(instance)
+			ph.ReclaimLeavingShards(instance)
 			p = ph.GeneratePlacement()
 		}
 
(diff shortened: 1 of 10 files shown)


placement/algo/sharded_helper.go
```diff
@@ -230,10 +260,15 @@ func (ph *placementHelper) targetLoadForInstance(id string) int {
 }
 
 func (ph *placementHelper) moveOneShard(from, to placement.Instance) bool {
-	for _, s := range from.Shards().All() {
-		if s.State() != shard.Leaving && ph.moveShard(s, from, to) {
-			return true
-		}
+	// Prefer to move a Unknown shard first, then Initializing, then Available.
```
xichen2020: Can you elaborate on the rationale behind this heuristic?

placement/algo/sharded_helper.go
```diff
@@ -230,10 +260,15 @@ func (ph *placementHelper) targetLoadForInstance(id string) int {
...
+	if ph.moveOneShardInState(from, to, shard.Unknown) {
+		return true
+	}
+	if ph.moveOneShardInState(from, to, shard.Initializing) {
+		return true
+	}
+	if ph.moveOneShardInState(from, to, shard.Available) {
+		return true
 	}
 	return false
```
xichen2020: nit: take it or leave it, but could be simplified to
```
return ph.moveOneShardInState(from, to, shard.Unknown) ||
  ph.moveOneShardInState(from, to, shard.Initializing) ||
  ph.moveOneShardInState(from, to, shard.Available)

File: placement/algo/sharded_helper.go [hunk 3]
```diff
@@ -354,14 +389,7 @@ func (ph *placementHelper) PlaceShards(
 		// NB(cw) when removing an adding instance that has not finished bootstrapping its
 		// Initializing shards, prefer to return those Initializing shards back to the leaving instance
 		// to reduce some bootstrapping work in the cluster.
-		if err := ph.returnInitializingShardsToSource(shardSet, from, candidates); err != nil {
-			return err
-		}
-		// prefer to distribute "some" of the load to other racks first
```
xichen2020: This is no longer needed?

placement/algo/sharded_helper.go
```diff
@@ -414,60 +447,17 @@ func (ph *placementHelper) returnInitializingShardsToSource(
 		}
 		sourceInstance, ok := candidateMap[sourceID]
 		if !ok {
-			return fmt.Errorf("could not find sourceID %s for instance %s, shard %d", sourceID, from.ID(), s.ID())
+			// NB(cw): This is not an error because the candidates are not
```
xichen2020: When this happens, these shards won't be able to be returned to the source instance, is that ok?

placement/algo/mirrored.go
```diff
@@ -205,12 +242,54 @@ func (a mirroredAlgorithm) ReplaceInstances(
...
+	}
+
+	for _, i := range p.Instances() {
+		if !i.IsInitializing() {
+			continue
+		}
+		if _, ok := ids[i.ID()]; !ok {
+			return false
+		}
+		delete(ids, i.ID())
```
xichen2020: nit: Take it or leave it, but use a `var totalInstances = len(instances)` and decrement that count every time there's a match is probably cheaper than deleting an entry from the map.

placement/algo/mirrored.go
```diff
@@ -205,12 +242,54 @@ func (a mirroredAlgorithm) ReplaceInstances(
...
+// allInitializing returns true when the given list of instances
+// matches all the leaving instances in the placement.
+func allLeaving(instances []placement.Instance, p placement.Placement) bool {
+	ids := make(map[string]struct{}, len(instances))
+	for _, i := range instances {
+		ids[i.ID()] = struct{}{}
+	}
+
+	for _, i := range p.Instances() {
+		if !i.IsLeaving() {
```
xichen2020: Can you refactor the common logic into a common `matchAllInState(...)` function and pass `func(i instance) bool { return i.IsLeaving() }` and  `func(i instance) bool { return i.IsInitializing() }` in these two functions? Reduces code duplication.

placement/algo/mirrored.go
```diff
@@ -91,6 +91,21 @@ func (a mirroredAlgorithm) RemoveInstances(
 		return nil, err
 	}
 
+	// If the instances being removed are all the initializing instances in the placement.
+	// We just need to return these shards back to their sources.
+	if allInitializing(instanceIDs, p) {
```
xichen2020: What happens if `instanceIDs` only contain a subset of instance ids that are initializing? What if it's a superset?

placement/algo/sharded_helper.go
```diff
@@ -48,27 +48,35 @@ const (
...
+	// Optimize rebalances the load distribution in the cluster.
+	Optimize(t optimizeType) error
 
-	// GeneratePlacement generates a placement
+	// GeneratePlacement generates a placement.
 	GeneratePlacement() placement.Placement
+
+	// CleanUpLeavingShards cleans up all the leaving shards on the given instance
+	// by pulling them back from the rest of the cluster.
+	CleanUpLeavingShards(instance placement.Instance)
```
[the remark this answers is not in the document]
cw9: sg

placement/algo/sharded_helper.go
```diff
@@ -48,27 +48,35 @@ const (
...
+
+	// CleanUpInitializingShards cleans up all the initializing shards on the given instance
+	// by returning them back to the original owners.
+	CleanUpInitializingShards(instance placement.Instance)
```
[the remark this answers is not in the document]
cw9: sg

placement/algo/sharded_helper.go
cw9: Modified the comment

placement/algo/sharded_helper.go
cw9: sg

placement/algo/sharded_helper.go
cw9: No longer needed, we use Optimize() to achieve better load distribution now.

placement/algo/sharded_helper.go
cw9: That's ok, if you pass in all the instances in the placement then this won't happen, for cases when you only want to move shards to a subset of instances by calling PlaceShards(), shards will be moved to those instances even though they are not the source instance for the Initializing shards.

placement/algo/mirrored.go
cw9: then the reverting won't be triggered, it will go through the normal Remove path.

placement/algo/mirrored.go
cw9: sg

placement/algo/sharded_helper.go
xichen2020: Sounds good.

PR #158, opened 2017-09-12
Add Clone to Shard, Instance and Placement interface

cw9:
move the Clone() functions out from placement.go

@xichen2020

### DIFF
--- a/placement/placement.go
+++ b/placement/placement.go
@@ -212,6 +212,16 @@ func (p *placement) Proto() (*placementpb.Placement, error) {
 	}, nil
 }
 
+func (p *placement) Clone() Placement {
+	return NewPlacement().
+		SetInstances(Instances(p.Instances()).Clone()).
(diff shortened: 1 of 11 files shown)


placement/placement.go
```diff
@@ -659,16 +629,15 @@ func markShardAvailable(p Placement, instanceID string, shardID uint32) (Placeme
 
 	sourceShards.Remove(shardID)
 	if sourceShards.NumShards() == 0 {
-		return ClonePlacement(p).
-			SetInstances(RemoveInstanceFromList(p.Instances(), sourceInstance.ID())), nil
+		return p.SetInstances(RemoveInstanceFromList(p.Instances(), sourceInstance.ID())), nil
```
xichen2020: No need to clone any more?

cw9: nope, the 2 functions that are calling this function did a clone already

PR #159, opened 2017-09-13
Expose ActiveIndex on Placements

cw9:
So I can access this function in statsdex
@xichen2020

### DIFF
--- a/placement/placement.go
+++ b/placement/placement.go
@@ -26,9 +26,10 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/m3db/m3cluster/generated/proto/placementpb"
 	"github.com/m3db/m3cluster/shard"
+
+	"github.com/golang/protobuf/proto"
 )
 
 var (
--- a/placement/placement.go
+++ b/placement/placement.go
@@ -228,7 +228,6 @@ type Placements []Placement
 
 // NewPlacementsFromProto creates a list of placements from proto.
 func NewPlacementsFromProto(p *placementpb.PlacementSnapshots) (Placements, error) {
-	// If pb is a typed nil pointer, we may get a valid nil pointer for p.
 	if p == nil {
 		return nil, errNilPlacementSnapshotsProto
 	}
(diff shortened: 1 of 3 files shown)


placement/placement.go
```diff
@@ -26,17 +26,20 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/golang/protobuf/proto"
```
xichen2020: Reorder imports?

placement/placement.go
```diff
@@ -56,10 +59,21 @@ func NewPlacement() Placement {
 }
 
 // NewPlacementFromProto creates a new placement from proto.
-func NewPlacementFromProto(p *placementpb.Placement) (Placement, error) {
+func NewPlacementFromProto(pb proto.Message) (Placement, error) {
```
xichen2020: Why do we need to pass in a `proto.Message` instead of `placementpb.Placement`?

placement/placement.go
```diff
@@ -226,7 +240,17 @@ func (p *placement) Clone() Placement {
 type Placements []Placement
 
 // NewPlacementsFromProto creates a list of placements from proto.
-func NewPlacementsFromProto(p *placementpb.PlacementSnapshots) (Placements, error) {
+func NewPlacementsFromProto(pb proto.Message) (Placements, error) {
```
xichen2020: Same here.

placement/placement.go
cw9: Was trying to save statsdex from knowing the exact type of proto for constructing a Placement, but on the other hand since we already expose the protobuf type on the Placement interface, I guess the abstraction does not give too much value now.

placement/placement.go
```diff
@@ -227,6 +228,7 @@ type Placements []Placement
 
 // NewPlacementsFromProto creates a list of placements from proto.
 func NewPlacementsFromProto(p *placementpb.PlacementSnapshots) (Placements, error) {
+	// If pb is a typed nil pointer, we may get a valid nil pointer for p.
```
xichen2020: Remove the comment?

PR #163, opened 2017-09-14
[util] Return watch from WatchAndUpdate functions

jeromefroe:
This PR returns the ValueWatches from the `WatchAndUpdate` functions so the caller can close them if they need to. Callers would then be able to ensure that the goroutines which watch for these updates will be closed.

cc @prateek @cw9

### DIFF
--- a/kv/util/util_test.go
+++ b/kv/util/util_test.go
@@ -58,7 +58,9 @@ func TestWatchAndUpdateBool(t *testing.T) {
 		defaultValue = false
 	)
 
-	err := WatchAndUpdateBool(store, "foo", &testConfig.v, &testConfig.RWMutex, defaultValue, nil)
+	watch, err := WatchAndUpdateBool(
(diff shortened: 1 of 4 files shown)


kv/util/util_test.go
```diff
@@ -222,6 +244,14 @@ func TestWatchAndUpdateInt64(t *testing.T) {
 			break
 		}
 	}
+
+	// Updates should not be applied after the watch is closed.
+	watch.Close()
```
prateek: could you ensure we're not leaking goroutines (use something like [leaktest](https://github.com/fortytw2/leaktest) )

kv/util/util_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/fortytw2/leaktest"
```
prateek: nit: import ordering

PR #165, opened 2017-09-15
Fix the bug where instance could reclaim all of its leaving shards

cw9:
In very rare case, the leaving shards could not be taken back due to rack conflicts.
For example, in a RF=2 case, instance a and b on rack1, instance c on rack2, instance d on rack3,
a and c owns shard 1, 

Remove a and c:
1, a give shard 1 to d
2, c give shard 1 to b

Add a and c back:
Now if a tries to get shard 1 back from d there will be a rack conflict because b(on the same rack with a) owns shard 1 as well.
This rack conflict can be resolved by having c take the shard back from b first, then a can take the shard back from d without rack conflict.

This diff introduces a loop that retries the shard reclaiming process as long as some instance are still making progress.

### DIFF
--- a/placement/algo/mirrored.go
+++ b/placement/algo/mirrored.go
@@ -93,17 +93,8 @@ func (a mirroredAlgorithm) RemoveInstances(
 
 	// If the instances being removed are all the initializing instances in the placement.
 	// We just need to return these shards back to their sources.
-	if allInitializing(instanceIDs, p) {
-		for _, id := range instanceIDs {
-			ph, instance, err := newRemoveInstanceHelper(p, id, a.opts)
-			if err != nil {
-				return nil, err
-			}
-			ph.ReturnInitializingShards(instance)
-			p = ph.GeneratePlacement()
-		}
-
-		return p, nil
+	if allInitializing(p, instanceIDs) {
+		return a.returnInitializingShards(p, instanceIDs)
 	}
 
 	p, err := placement.MarkAllShardsAsAvailable(p)
@@ -157,17 +148,8 @@ func (a mirroredAlgorithm) AddInstances(
 
 	// If the instances being added are all the leaving instances in the placement.
 	// We just need to get their shards back.
-	if allLeaving(addingInstances, p) {
-		for _, instance := range addingInstances {
-			ph, instance, err := newAddInstanceHelper(p, instance, a.opts)
-			if err != nil {
-				return nil, err
-			}
-			ph.ReclaimLeavingShards(instance)
-			p = ph.GeneratePlacement()
-		}
-
-		return p, nil
+	if allLeaving(p, addingInstances) {
+		return a.reclaimLeavingShards(p, addingInstances)
 	}
 
 	p, err := placement.MarkAllShardsAsAvailable(p)
@@ -195,7 +177,7 @@ func (a mirroredAlgorithm) AddInstances(
 			return nil, fmt.Errorf("shard set id %d already exist in current placement", instance.ShardSetID())
 		}
 
-		ph, instance, err := newAddInstanceHelper(mirrorPlacement, instance, a.opts)
+		ph, instance, err := newAddEmptyInstanceHelper(mirrorPlacement, instance, a.opts)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


prateek:
@cw9 I think we should take this chance to write some [property tests](https://github.com/leanovate/gopter) for `m3cluster`. We've been adding to `m3db` recently. They're great at catching these kinds of issues.

placement/algo/sharded_helper.go
```diff
@@ -126,7 +131,19 @@ func newAddInstanceHelper(
 	}
 
 	return newHelper(p, p.ReplicaFactor(), opts), instanceInPlacement, nil
+}
 
+func newAddInstanceHelper(
```
xichen2020: This looks quite similar to `newAddEmptyInstanceHelper`, what's the difference between the two? When would we use one or the other?

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
 
 	return len(instanceIDs) == 0
 }
+
+func (a mirroredAlgorithm) returnInitializingShards(
```
xichen2020: Can you add some comments briefly explaining the algorithm?

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
...
+	p placement.Placement,
+	instanceIDs []string,
+) (placement.Placement, error) {
+	shouldContinue := true
```
xichen2020: nit: take it or leave it, but `progressed` or `madeProgress` might be easier to understand and more specific than `shouldContinue`.

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
...
+	for shouldContinue {
```
xichen2020: nit: might be easier to understand to write it as (also this narrows the scope of `madeProgress`)
```
for {
  madeProgress := false
  for ... {
   }
   if !madeProgress {
      break
   }
}
```

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
...
+		numInitializingShards := instance.Shards().NumShardsForState(shard.Initializing)
+		if numInitializingShards != 0 {
+			return nil, fmt.Errorf("there are %d initializing shards could not be returned for instance %s", numInitializingShards, id)
+		}
+	}
+
+	return p, nil
+}
+
+func (a mirroredAlgorithm) reclaimLeavingShards(
```
xichen2020: Can we add some comments here explaining the algorithm?

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
...
+	p placement.Placement,
+	addingInstances []placement.Instance,
+) (placement.Placement, error) {
+	shouldContinue := true
```
xichen2020: Same nit w.r.t. `shouldContinue`, see above.

placement/algo/mirrored.go
```diff
@@ -285,6 +267,83 @@ func allInstancesInState(
...
+				shouldContinue = true
+			}
+			p = ph.GeneratePlacement()
+		}
+	}
+
+	for _, instance := range addingInstances {
+		instance, ok := p.Instance(instance.ID())
+		if !ok {
+			return nil, fmt.Errorf("could not find instance %s in placement after reclaiming leaving shards", instance.ID())
```
xichen2020: Probably should do
```
instanceID := instance.ID()
instance, ok := p.Instance(InstanceID)
if !ok {
   return nil, fmt.Errorf("...", instanceID)
}
...
```
? Otherwise it might panic.

placement/algo/sharded_helper.go
cw9: newAddEmptyInstanceHelper only adds instance that is not in the placement or is leaving in the placement, basically has no load on it.

placement/algo/mirrored.go
cw9: sure

placement/algo/mirrored.go
cw9: yeah good catch

cw9:
@prateek interesting, will look into that

PR #169, opened 2017-09-18
[util] Add functions to update values atomically

jeromefroe:
This PR adds new utility functions to update their properties atomically instead of acquiring locks. It also does some minor refactoring now that we have functions for updating with a lock and updating atomically. 

cc @xichen2020 @cw9 @prateek


schallert:
LGTM

PR #170, opened 2017-09-19
basic plumbing to allow setting client tls for kv clusters

pjjw:
with this change we can distribute certs with langley and use them for auth, rather than exposing a bare etcd cluster to the company. can use a read-only cert in the proxy to avoid anyone accidentally trying to write stuff to KV after looking at the collector.

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -245,7 +245,7 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 
 func newClient(endpoints []string, certfile string, keyfile string, ca string) (*clientv3.Client, error) {
 	tlscfg := &tls.Config{}
-	if certfile != "" && keyfile != "" {
+	if certfile != "" {
 		cert, err := tls.LoadX509KeyPair(certfile, keyfile)
 		if err != nil {
 			return nil, err
@@ -261,8 +261,7 @@ func newClient(endpoints []string, certfile string, keyfile string, ca string) (
 				return nil, err
 			}
 			caPool := x509.NewCertPool()
-			ok := caPool.AppendCertsFromPEM(caCert)
-			if !ok {
+			if ok := caPool.AppendCertsFromPEM(caCert); !ok {
 				return nil, fmt.Errorf("can't read PEM-formatted certificates from file %s as root CA pool", ca)
 			}
 			tlscfg.RootCAs = caPool
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -252,7 +252,7 @@ func newClient(endpoints []string, certfile string, keyfile string, ca string) (
 		}
 		tlscfg = &tls.Config{
 			MinVersion:         tls.VersionTLS12,
-			InsecureSkipVerify: true,
+			InsecureSkipVerify: false,
 			Certificates:       []tls.Certificate{cert},
 		}
 		if ca != "" {
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -26,7 +26,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -40,6 +39,7 @@ import (
 	etcdheartbeat "github.com/m3db/m3cluster/services/heartbeat/etcd"
 	"github.com/m3db/m3cluster/services/leader"
 	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/uber-go/tally"
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -21,8 +21,12 @@
 package etcd
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -36,7 +40,6 @@ import (
 	etcdheartbeat "github.com/m3db/m3cluster/services/heartbeat/etcd"
 	"github.com/m3db/m3cluster/services/leader"
 	"github.com/m3db/m3x/instrument"
-	"github.com/m3db/m3x/log"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/uber-go/tally"
@@ -53,7 +56,7 @@ const (
 
 var errInvalidNamespace = errors.New("invalid namespace")
 
-type newClientFn func(endpoints []string) (*clientv3.Client, error)
+type newClientFn func(endpoints []string, cert string, key string, ca string) (*clientv3.Client, error)
 
 type cacheFileForZoneFn func(zone string) etcdkv.CacheFileFn
 
@@ -231,7 +234,7 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 		return nil, fmt.Errorf("no etcd cluster found for zone %s", zone)
 	}
 
-	cli, err := c.newFn(cluster.Endpoints())
+	cli, err := c.newFn(cluster.Endpoints(), cluster.Cert(), cluster.Key(), cluster.CA())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


client/etcd/config_test.go
```diff
@@ -45,6 +45,15 @@ const testConfig = `
...
               - etcd4:2379
+          cert: foo.crt.pem
+          key: foo.key.pem
+        - zone: z3
+          endpoints:
+              - etcd5:2379
+              - etcd6:2379
+          cert: foo.crt.pem
+          key: foo.key.pem
+		  ca: foo_ca.pem
```
cw9: nit: fix indentation

client/etcd/client_test.go
```diff
@@ -274,7 +274,7 @@ func testNewETCDFn(t *testing.T) (newClientFn, func()) {
 	ecluster := integration.NewClusterV3(t, &integration.ClusterConfig{Size: 1})
 	ec := ecluster.RandClient()
 
-	newFn := func(endpoints []string) (*clientv3.Client, error) {
+	newFn := func(endpoints []string, cert string, key string) (*clientv3.Client, error) {
```
cw9: seems you also need to add "ca string" to the arguments

client/etcd/client.go
```diff
@@ -240,8 +243,32 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 	return cli, nil
 }
 
-func newClient(endpoints []string) (*clientv3.Client, error) {
-	return clientv3.New(clientv3.Config{Endpoints: endpoints})
+func newClient(endpoints []string, certfile string, keyfile string, ca string) (*clientv3.Client, error) {
+	tlscfg := &tls.Config{}
+	if certfile != "" && keyfile != "" {
```
cw9: should we return error when only one of certifile and keyfile is available?

client/etcd/client.go
```diff
@@ -240,8 +243,32 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
...
+			InsecureSkipVerify: true,
+			Certificates:       []tls.Certificate{cert},
+		}
+		if ca != "" {
+			caCert, err := ioutil.ReadFile(ca)
+			if err != nil {
+				return nil, err
+			}
+			caPool := x509.NewCertPool()
+			ok := caPool.AppendCertsFromPEM(caCert)
```
cw9: nit: `if ok := caPool.AppendCertsFromPEM(caCert); !ok {...}`

client/etcd/client.go
pjjw: i'll just check certfile only for zero-value, then the error should be handled by the failure in tls.LoadX509KeyPair.

pjjw:
you're gonna have to help me figure out why this dang thing isn't passing CI tomorrow.

client/etcd/client.go
```diff
@@ -36,7 +40,6 @@ import (
 	etcdheartbeat "github.com/m3db/m3cluster/services/heartbeat/etcd"
 	"github.com/m3db/m3cluster/services/leader"
 	"github.com/m3db/m3x/instrument"
-	"github.com/m3db/m3x/log"
```
jeromefroe: Seems `goimports` might have removed this package (`github.com/m3db/m3x/log`) in favor or `log`. I think you'll need to switch it back though since everything things will expect the former. (This is one of those magic things VS Code runs for you by default that are usually pretty helpful but can go awry at times.)

pjjw:
lol wow. that's pretty shitty of it!

pjjw:
more though that there's a submodule called '.ci' that doesn't seem to refer to any remote.

jeromefroe:
@pjjw yea that's the CI stuff, need to run `git submodule update --init --recursive` to pull it down after a fresh clone

client/etcd/config_test.go
```diff
@@ -61,8 +70,20 @@ func TestConfig(t *testing.T) {
...
+		ClusterConfig{Zone: "z1", Endpoints: []string{"etcd1:2379", "etcd2:2379"}},
+		ClusterConfig{
+			Zone:      "z2",
+			Endpoints: []string{"etcd3:2379", "etcd4:2379"},
+			Cert:      "foo.crt.pem",
+			Key:       "foo.key.pem",
+		},
+		ClusterConfig{
+			Zone:      "z3",
+			Endpoints: []string{"etcd3:2379", "etcd4:2379"},
```
jeromefroe: Should be `etcd5` and `etcd6` respectively.

client/etcd/config.go
```diff
@@ -30,6 +30,9 @@ import (
 type ClusterConfig struct {
 	Zone      string   `yaml:"zone"`
 	Endpoints []string `yaml:"endpoints"`
+	Cert      string   `yaml:"cert",omitempty`
```
prateek: nit: group these fields into a struct and add a suffix to indicate these string refer to a path. E.g. https://github.com/m3db/m3em/blob/master/services/m3em_agent/config/config.go#L67-L72

client/etcd/options.go
```diff
@@ -61,6 +61,13 @@ type Cluster interface {
 
 	Endpoints() []string
 	SetEndpoints(endpoints []string) Cluster
+
+	Cert() string
```
prateek: nit: `CertPath` and so on ..

glide.yaml
```diff
@@ -25,8 +25,8 @@ import:
   version: ^3.0.0 # ie >= 3.0.0, < 4.0.0
 - package: github.com/apache/thrift
   version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
-- package: github.com/uber-go/atomic
-  version: e682c1008ac17bf26d2e4b5ad6cdd08520ed0b22
+- package: go.uber.org/atomic
+  version: ^1
```
prateek: nice

client/etcd/client.go
```diff
@@ -240,8 +243,31 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 	return cli, nil
 }
 
-func newClient(endpoints []string) (*clientv3.Client, error) {
-	return clientv3.New(clientv3.Config{Endpoints: endpoints})
+func newClient(endpoints []string, certfile string, keyfile string, ca string) (*clientv3.Client, error) {
+	tlscfg := &tls.Config{}
+	if certfile != "" {
```
prateek: suggestion(take/leave): group the cert specific code into a function on the sub-type for the config. 

i.e. 
```
type TLSConfiguration struct {
  KeyPath string `yaml...`
  CrtPath ... 
  CACrtPath .. 
} 

func (t TLSConfiguration) NewTLSConfig() (*tls.Config, error) {
  ... 
}
```

placement/algo/sharded.go
```diff
@@ -168,11 +168,7 @@ func (a rackAwarePlacementAlgorithm) ReplaceInstances(
 			return nil, err
 		}
 		load := loadOnInstance(leavingInstance)
-		if load == 0 {
-			result, _, err := addInstanceToPlacement(ph.GeneratePlacement(), leavingInstance, withShards)
-			return result, err
-		}
-		if !a.opts.AllowPartialReplace() {
+		if load != 0 && !a.opts.AllowPartialReplace() {
```
prateek: the changes in this file, glide, and `kv/util/...` seem unrelated to TLS changes. Any reason to group them in this PR?

pjjw: rebased from master to try and figure out why CI was barfing, i'll undo the rebase

client/etcd/client.go
pjjw: I'd agree with you if there were any other logic in the newClient function beforehand, but as is there's nothing, so all we'd do is add another layer of if err != nil boilerplate around the whole thing. Would prefer to leave as-is.

cw9: I'd +1 to Prateek's point, the tls logic is pretty isolated, if we ever need to change the tls logic we just need to add fields to the TLSConfig struct without having to add another field to this function.

client/etcd/config.go
cw9: nit: probably better to do `TLS *TLSConfiguration` to indicate it's optional

client/etcd/client.go
pjjw: Like I said, there's no logic to separate- before the TLS logic this function was just a return statement. There's nothing else having to do with TLS that etcd cares about- almost all the other fields available in the tls.Config object are constrained by GRPC's restrictions on TLS, and we'd have to make the get/set fields in the tls config object anyhow, unless I'm misunderstanding your suggestion.

PR #171, opened 2017-09-19
Disallow marking shards as available before cutover, Disallow reverting after cutover

cw9:
Moved MarkShardAvailable to algo interface
Check current time before marking shards as available


placement/algo/mirrored.go
```diff
@@ -197,16 +206,25 @@ func (a mirroredAlgorithm) ReplaceInstances(
...
 		return nil, fmt.Errorf("could not replace %d instances with %d instances for mirrored replace", len(leavingInstanceIDs), len(addingInstances))
 	}
 
-	p, err := placement.MarkAllShardsAsAvailable(p)
-	if err != nil {
+	var (
+		nowNanos               = a.opts.NowFn()().UnixNano()
+		shouldCheckCutoverTime = true
+	)
+	if allLeaving(p, addingInstances, nowNanos) && allInitializing(p, leavingInstanceIDs, nowNanos) {
```
jeromefroe: This seems backwards to me, how come we don't want to check if the leaving instances are leaving and the added instances are initializing? i.e

`if allLeaving(p, leavingInstanceIDs, nowNanos) && allInitializing(p, addingInstances, nowNanos) {`

placement/algo/mirrored.go
```diff
@@ -224,45 +242,68 @@ func (a mirroredAlgorithm) ReplaceInstances(
...
+	for _, instance := range p.Instances() {
+		if !instanceCheck(instance, forEachShardFn) {
 			continue
 		}
-		if _, ok := instanceIDs[i.ID()]; !ok {
+		if _, ok := instanceIDs[instance.ID()]; !ok {
 			return false
 		}
-		delete(instanceIDs, i.ID())
+		delete(instanceIDs, instance.ID())
```
jeromefroe: nit: perhaps it would be preferable to not modify `instanceIDs` in case we ever want to reuse it after calling `allInstancesInState`? Seems like we could get the same effect by just incrementing a counter each time through the loop instead of deleting from `instanceIDs` and the checking that the counter equals the length of `instanceIDs`?

placement/algo/sharded_helper.go
```diff
@@ -738,3 +738,80 @@ func newShards(shardIDs []uint32) []shard.Shard {
 	}
 	return r
 }
+
+func removeInstanceFromList(list []placement.Instance, instanceID string) []placement.Instance {
+	for i, instance := range list {
+		if instance.ID() == instanceID {
+			last := len(list) - 1
+			list[i], list[last] = list[last], list[i]
```
jeromefroe: nit: do we need the swap here? Seems it might be simpler to do:

```
last := len(list)-1
list[i] = list[last]
return list[:last]
```

placement/algo/sharded_helper.go
```diff
@@ -738,3 +738,80 @@ func newShards(shardIDs []uint32) []shard.Shard {
...
+	}
+
+	if shouldCheckCutoverTime && s.CutoverNanos() > nowNanos {
+		return nil, fmt.Errorf("could not mark shard %d as available, it's cutover time %d is scheduled after now %d", s.ID(), s.CutoverNanos(), nowNanos)
+	}
+
+	sourceID := s.SourceID()
+	shards.Add(shard.NewShard(shardID).SetState(shard.Available))
+
+	// there could be no source for cases like initial placement
```
jeromefroe: nit: start comment with capital letter and end with period perhaps?

placement/algo/sharded_test.go
```diff
@@ -1174,7 +1203,7 @@ func validateDistribution(t *testing.T, p placement.Placement, expectPeakOverAvg
 				testCase, i.ID(), instanceOverTarget, load, targetLoad))
 		}
 	}
-	assert.Equal(t, total, p.ReplicaFactor()*p.NumShards(), fmt.Sprintf("Wrong total shards: expecting %v, but got %v", p.ReplicaFactor()*p.NumShards(), total))
+	assert.Equal(t, p.ReplicaFactor()*p.NumShards(), total, fmt.Sprintf("Wrong total shards: expecting %v, but got %v", p.ReplicaFactor()*p.NumShards(), total))
```
jeromefroe: given that we are requiring that all shards be marked as available, as indicated by the use of must in the function, perhaps we want to use `require` instead of `assert` here? With `assert` the tests will continue even if condition is not true, but with `require` they will stop immediately.

placement/algo/mirrored.go
cw9: this is exactly what we want, we often need to add instances that are leaving in the placement, for example you have 10 instances in normal (Available) state, and you removed 2 of them, the 2 instances will become "leaving" in the placement, then it's totally ok if we want to add the 2 "leaving" instances back to the placement

placement/algo/mirrored.go
cw9: the instanceIDs are created purely to be used in this function, I could do the inc/dec counter for performance gain, but seems like an overkill

placement/algo/sharded_helper.go
cw9: yeah, could save the swap

placement/algo/sharded_helper.go
cw9: will do

placement/algo/sharded_test.go
cw9: we use assert instead of require in many places in this test so we can capture more errors in one test, I felt it helped investigating. Can change some NoError checks to require

placement/algo/sharded_helper.go
```diff
@@ -738,3 +738,80 @@ func newShards(shardIDs []uint32) []shard.Shard {
...
+	sourceShard, exist := sourceShards.Shard(shardID)
+	if !exist {
+		return nil, fmt.Errorf("shard %d does not exist in source instance %s", shardID, sourceID)
+	}
+
+	if sourceShard.State() != shard.Leaving {
+		return nil, fmt.Errorf("shard %d is not leaving instance %s", shardID, sourceID)
+	}
+
+	sourceShards.Remove(shardID)
```
xichen2020: Do we need to consider cutoff times here? In particular, if a shard is cut off at 2pm, but doesn't flush all its data until 2:50pm, we probably don't want to remove the leaving shard from the placement, which causes the shard to be closed immediately on the instance. 

Perhaps we should set `shardCutoverFn` and `shardCutoffFn` on the placement options, and use the two functions to evaluate when a shard can be marked as available / removed, where the `shardCutoverFn` compares the cutover time against now, and the `shardCutoffFn` compares the cutoff + maxAggregationWindow against now.

placement/algo/mirrored.go
```diff
@@ -197,16 +206,25 @@ func (a mirroredAlgorithm) ReplaceInstances(
...
 
-	p, err := placement.MarkAllShardsAsAvailable(p)
-	if err != nil {
+	var (
+		nowNanos               = a.opts.NowFn()().UnixNano()
+		shouldCheckCutoverTime = true
+	)
+	if allLeaving(p, addingInstances, nowNanos) && allInitializing(p, leavingInstanceIDs, nowNanos) {
+		// Allow marking shards as available without checking cutover time in a reverting case.
+		shouldCheckCutoverTime = false
```
xichen2020: Hmm if we are reverting a replace (i.e., removing the initializing instance and adding back the leaving instance before the shard transition happens), and we mark shards available below, wouldn't that mark the initializing shards as available and remove the leaving shards, which essentially marks the replace as complete?

placement/algo/sharded_helper.go
```diff
@@ -738,3 +738,80 @@ func newShards(shardIDs []uint32) []shard.Shard {
 	}
 	return r
 }
+
+func removeInstanceFromList(list []placement.Instance, instanceID string) []placement.Instance {
```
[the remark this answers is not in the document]
cw9: sure

placement/algo/sharded_helper.go
cw9: yeah good point, could compare the cutoff time + max window as well

placement/algo/sharded_helper.go
```diff
@@ -738,3 +738,80 @@ func newShards(shardIDs []uint32) []shard.Shard {
...
+	}
+
+	sourceShards.Remove(shardID)
+	if sourceShards.NumShards() == 0 {
+		return p.SetInstances(removeInstanceFromList(p.Instances(), sourceInstance.ID())), nil
+	}
+	return p, nil
+}
+
+func markAllShardsAvailable(p placement.Placement, shouldCheckCutoverTime bool, nowNanos int64) (placement.Placement, error) {
```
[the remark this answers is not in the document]
cw9: We do, this is mostly used in BuildInititialPlacement(), all the shard in the initial placement will all have cutover time in the future rather than 0.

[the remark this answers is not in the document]
cw9: it helps with testing, I'll probably pass in a placement options here now that we are separating cutover compare and cutoff compare

placement/algo/mirrored.go
cw9: It marks the previous one as complete and then do another replace for the revert

placement/algo/sharded_helper.go
xichen2020: Probably don't want to introduce the concept of "max window" here though, which is specific to the agg tier placement generation. I'd just pass in the two functions and provide them in the tool for generating agg tier placement.

placement/algo/mirrored.go
xichen2020: We do that before the shard transition happens? I thought if the shard transition hasn't happened for a replace and we perform a revert, we should just return the initializing shards and remove the leaving shards and there should be no leaving/initializing shards in the final placement. Is that not true?

placement/algo/sharded_helper.go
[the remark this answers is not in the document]
xichen2020: Is that necessary? I'd imagine the initial placement is for setting up the agg tier for some initial testing, so might be better to get feedback sooner than waiting till the cutover time. Not feeling strongly about it either way though.

placement/algo/sharded_helper.go
cw9: ? I just passed the 2 validation functions to the function? the m3admin tooling will be creating the cutoff validation function with max window taken into consideration

placement/algo/sharded_helper.go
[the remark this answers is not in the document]
cw9: This means the generated initial placement will have all shards marked as available, the markShardAsAvailable was done in placement service and I moved that to the algo in this diff

placement/algo/mirrored.go
cw9: We are only reverting for add and remove right now, we could do the same for replace but for some reason we are not doing that, I can't recall why now, but adding the same reverting for replace is simple, I could do that in a follow up diff if that's what we want.

placement/algo/sharded_helper.go
xichen2020: Right yeah that's what I meant, not sure whether you were doing that so figured I should raise it, sounds good then.

placement/algo/sharded_helper.go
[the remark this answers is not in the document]
xichen2020: Oh I see cool sounds good.

placement/algo/mirrored.go
xichen2020: Yeah I think that's what we'd want for replaces too. Perhaps just address it in this PR while you are there?

xichen2020:
Also seems `TestWatchAndAtomicUpdateBool` is timing out?

placement/algo/mirrored.go
cw9: Sure, sounds good

cw9:
Seems like a previous fix on TestWatchAndAtomicUpdateBool got reverted, fixed it again

placement/algo/mirrored.go
```diff
@@ -217,16 +217,13 @@ func (a mirroredAlgorithm) ReplaceInstances(
 		return nil, fmt.Errorf("could not replace %d instances with %d instances for mirrored replace", len(leavingInstanceIDs), len(addingInstances))
 	}
 
-	var (
-		nowNanos = a.opts.NowFn()().UnixNano()
-		opts     = a.opts
-	)
+	nowNanos := a.opts.NowFn()().UnixNano()
 	if allLeaving(p, addingInstances, nowNanos) && allInitializing(p, leavingInstanceIDs, nowNanos) {
 		// Allow marking shards as available without checking cutover time in a reverting case.
```
xichen2020: Update the comment here as well?

PR #172, opened 2017-09-25
Use nil tls config by default

cw9:
Currently we use an empty tls config by default, which causes the auth to fail when no certificate provided in config, should use nil for the no-auth case

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -244,7 +244,7 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 }
 
 func newClient(endpoints []string, certfile string, keyfile string, ca string) (*clientv3.Client, error) {
-	tlscfg := &tls.Config{}
+	var tlscfg *tls.Config
 	if certfile != "" {
 		cert, err := tls.LoadX509KeyPair(certfile, keyfile)
 		if err != nil {


client/etcd/client.go
```diff
@@ -244,7 +244,7 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
 }
 
 func newClient(endpoints []string, certfile string, keyfile string, ca string) (*clientv3.Client, error) {
-	tlscfg := &tls.Config{}
+	var tlscfg *tls.Config
```
prateek: While you're here, could you make the other changes we discussed in  #170

cw9: Yeah would prefer to do it in next diff, I need to pull latest m3cluster in and this bug is blocking it.

prateek: Cool, makes sense

client/etcd/client.go
```diff
@@ -244,7 +244,7 @@ func (c *csclient) etcdClientGen(zone string) (*clientv3.Client, error) {
...
 	if certfile != "" {
```
xichen2020: Would suggest moving this block to an `initTLSConfig` method since that's pretty isolated from the rest of the stuff.

cw9: yeah that's the refactoring we were talking about, I'll be refactoring this part later.

PR #173, opened 2017-09-25
Refactor TLS configuration

cw9:
Refactor the logic around auth to keep the tls related logic isolated from etcd cluster logic

### DIFF
--- a/client/etcd/config.go
+++ b/client/etcd/config.go
@@ -28,23 +28,23 @@ import (
 
 // ClusterConfig is the config for a zoned etcd cluster.
 type ClusterConfig struct {
-	Zone      string     `yaml:"zone"`
-	Endpoints []string   `yaml:"endpoints"`
-	Auth      AuthConfig `yaml:"auth"`
+	Zone      string    `yaml:"zone"`
+	Endpoints []string  `yaml:"endpoints"`
+	TLS       TLSConfig `yaml:"auth"`
 }
 
-// AuthConfig is the config for authentication.
-type AuthConfig struct {
-	Cert string `yaml:"cert"`
-	Key  string `yaml:"key"`
-	CA   string `yaml:"ca"`
+// TLSConfig is the config for TLS.
+type TLSConfig struct {
+	CrtPath   string `yaml:"crtPath"`
+	CACrtPath string `yaml:"caCrtPath"`
+	KeyPath   string `yaml:"keyPath"`
 }
 
-func (c AuthConfig) newOptions() AuthOptions {
-	return NewAuthOptions().
-		SetCert(c.Cert).
-		SetKey(c.Key).
-		SetCA(c.CA)
+func (c TLSConfig) newOptions() TLSOptions {
+	return NewTLSOptions().
+		SetCrtPath(c.CrtPath).
+		SetKeyPath(c.KeyPath).
+		SetCACrtPath(c.CACrtPath)
 }
 
 // Configuration is for config service client.
@@ -79,7 +79,7 @@ func (cfg Configuration) etcdClusters() []Cluster {
 		res[i] = NewCluster().
 			SetZone(c.Zone).
 			SetEndpoints(c.Endpoints).
-			SetAuthOptions(c.Auth.newOptions())
+			SetTLSOptions(c.TLS.newOptions())
 	}
 
 	return res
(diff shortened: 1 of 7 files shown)


client/etcd/config.go
```diff
@@ -26,16 +26,28 @@ import (
...
-	CA        string   `yaml:"ca",omitempty`
+	Zone      string     `yaml:"zone"`
+	Endpoints []string   `yaml:"endpoints"`
+	Auth      AuthConfig `yaml:"auth"`
 }
 
-// Configuration is for config service client
+// AuthConfig is the config for authentication.
+type AuthConfig struct {
+	Cert string `yaml:"cert"`
```
prateek: A few naming nits:

- rename to TLSConfig
- change var names to reflect what they are (path) and the type of file (crt/key).

Something like this:

```
type TLSConfiguration struct {
  KeyPath string `yaml...`
  CrtPath ... 
  CACrtPath .. 
}```

File: client/etcd/config.go [hunk 2]
```diff
@@ -26,16 +26,28 @@ import (
...
+// ClusterConfig is the config for a zoned etcd cluster.
 type ClusterConfig struct {
-	Zone      string   `yaml:"zone"`
-	Endpoints []string `yaml:"endpoints"`
-	Cert      string   `yaml:"cert",omitempty`
-	Key       string   `yaml:"key",omitempty`
-	CA        string   `yaml:"ca",omitempty`
+	Zone      string     `yaml:"zone"`
+	Endpoints []string   `yaml:"endpoints"`
+	Auth      AuthConfig `yaml:"auth"`
```
prateek: maybe make this type `*AuthConfig` to allow it to be nullable?

client/etcd/options.go
```diff
@@ -21,56 +21,80 @@
...
-	SetServiceDiscoveryConfig(cfg etcdsd.Configuration) Options
+// NewAuthOptions creates a set of Auth Options.
+func NewAuthOptions() AuthOptions {
+	return authOptions{}
+}
 
-	Clusters() []Cluster
-	SetClusters(clusters []Cluster) Options
-	ClusterForZone(z string) (Cluster, bool)
+type authOptions struct {
```
prateek: nit: same rename suggestions as above

client/etcd/config.go
cw9: The default struct can create a valid default option and saves a nil check, which is easier to use. Gonna keep it like this.

client/etcd/config.go
cw9: sg

client/etcd/options.go
cw9: sg

client/etcd/options.go
```diff
@@ -68,30 +68,29 @@ func (o tlsOptions) SetCACrtPath(ca string) TLSOptions {
 }
 
 func (o tlsOptions) Config() (*tls.Config, error) {
-	var tlscfg *tls.Config
 	if o.cert != "" {
```
prateek: minor nit: could you go the other route - 

```
if o.cert == "" {
  return nil, nil 
}
// rest of the code
```

cw9: true, more concise that way

client/etcd/config.go
```diff
@@ -26,16 +26,32 @@ import (
...
 
-// Configuration is for config service client
+// TLSConfig is the config for TLS.
+type TLSConfig struct {
+	CrtPath   string `yaml:"crtPath"`
+	CACrtPath string `yaml:"caCrtPath"`
+	KeyPath   string `yaml:"keyPath"`
+}
+
+func newTLSOptions(c *TLSConfig) TLSOptions {
```
xichen2020: nit: Convention is to do
```
func (c *TLSConfig) NewTLSOptions() {
...
}
```
and do the nil check in `NewOptions`.

client/etcd/types.go
```diff
@@ -0,0 +1,81 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package etcd
+
+import (
+	"crypto/tls"
+
+	etcdsd "github.com/m3db/m3cluster/services/client/etcd"
```
prateek: nit: import ordering

PR #174, opened 2017-09-26
Create KV store with custom zone and env

cw9:
@xichen2020 @jeromefroe

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -40,8 +40,8 @@ type Client interface {
 	Txn() (kv.TxnStore, error)
 
 	// Store returns access to the distributed configuration store with a namespace.
-	Store(namespace string) (kv.Store, error)
+	Store(opts kv.Options) (kv.Store, error)
 
 	// TxnStore returns access to the transaction store with a namespace.
-	TxnStore(namespace string) (kv.TxnStore, error)
+	TxnStore(opts kv.Options) (kv.TxnStore, error)
 }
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -40,8 +40,8 @@ type Client interface {
 	Txn() (kv.TxnStore, error)
 
 	// Store returns access to the distributed configuration store with a namespace.
-	Store(namespace string) (kv.Store, error)
+	Store(opts kv.Options) (kv.Store, error)
```
jeromefroe: How would you feel about instead of passing `kv.Options` in we created a `Prefixer` interface which looks like:

```
type Prefixer interface {
    Prefix() string
}
```

Then we can have a standard implementation which puts the zone, environment, and namespace together, but can also supports other kinds of prefixes where appropriate?

cw9: I'm hesitate to go down that route unless we have to, cause we use zone to create cluster rather than putting zone in the key. 
Unless one day we start to look at the prefix (or namespace) and find the cluster for that prefix and then look up the key in the cluster, we can't get away with only a prefix provided.

client/etcd/client.go
```diff
@@ -109,23 +109,28 @@ func (c *csclient) KV() (kv.Store, error) {
 
 func (c *csclient) Txn() (kv.TxnStore, error) {
 	c.txnOnce.Do(func() {
-		c.txn, c.txnErr = c.createTxnStore(kvPrefix)
+		c.txn, c.txnErr = c.createTxnStore(
+			kv.NewOptions().
+				SetNamespace(kvPrefix).
+				SetZone(c.opts.Zone()).
```
jeromefroe: I'm not sure we want to use the zone from the options here as that is the zone that determines where the keys are stored but the zone here should indicate which zone the keys are for. Would prefer a complete separation of the two so they aren't confused.

cw9: I'm ok with taking the zone option out now and figure out a solution later when we have concrete use cases.

PR #177, opened 2017-09-29
Add configuration for kv options

cw9:
@xichen2020 @jeromefroe

### DIFF
--- a/kv/config.go
+++ b/kv/config.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


kv/config.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
jeromefroe: nit: 2017?

kv/config.go
```diff
@@ -0,0 +1,32 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package kv
+
+// Configuration configs the kv options.
```
jeromefroe: nit: s/configs the/is the config for

kv/config.go
```diff
@@ -0,0 +1,32 @@
...
+type Configuration struct {
+	Environment string `yaml:"environment" validate:"nonzero"`
+	Namespace   string `yaml:"namespace" validate:"nonzero"`
```
xichen2020: Does the config always require a non-empty namespace?

kv/config.go
cw9: sg

kv/config.go
cw9: sg

kv/config.go
cw9: yeah, if you override, you'd better provide override for both fields

PR #180, opened 2017-10-06
Use default value when no override is provided in kv options

cw9:
@xichen2020


client/etcd/client.go
```diff
@@ -125,10 +125,8 @@ func (c *csclient) Store(opts kv.Options) (kv.Store, error) {
 }
 
 func (c *csclient) TxnStore(opts kv.Options) (kv.TxnStore, error) {
-	if err := opts.Validate(); err != nil {
-		return nil, err
-	}
-	if err := validateTopLevelNamespace(opts.Namespace()); err != nil {
+	opts, err := c.validateOverrideKVOptions(opts)
```
xichen2020: nit: `sanitizeOptions`

PR #181, opened 2017-10-09
Make the logger of a KV store configurable

xichen2020:
cc @cw9 @jeromefroe 

This PR makes the logger of a KV store configurable.

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -152,19 +152,24 @@ func (c *csclient) createTxnStore(opts kv.Options) (kv.TxnStore, error) {
 	if err := opts.Validate(); err != nil {
 		return nil, err
 	}
-	return c.txnGen(c.opts.Zone(), c.cacheFileFn(), opts.Namespace(), opts.Environment())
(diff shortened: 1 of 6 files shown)


client/etcd/client.go
```diff
@@ -152,19 +152,24 @@ func (c *csclient) createTxnStore(opts kv.Options) (kv.TxnStore, error) {
...
+func (c *csclient) newkvOptions(
+	zone string,
+	cacheFileFn cacheFileForZoneFn,
+	logger log.Logger,
+	namespaces ...string,
+) etcdkv.Options {
 	opts := etcdkv.NewOptions().
 		SetInstrumentsOptions(instrument.NewOptions().
-			SetLogger(c.logger).
+			SetLogger(logger).
```
jeromefroe: nit: would this mean that you have to explicitly set the logger in the options to receive updates? Maybe we should check if it's `nil` and fall back to `c.logger` if so?

xichen2020: If the logger is not set, it falls back to using the client's logger. The logger set in the options is more like an override.

PR #183, opened 2017-10-19
Expose endpoints for inntegration embedded KV server

robskillington:
This is required because I need to spin up a KV server and spin up M3DB using YAML to point to the KV server, i.e. can't use the call to `ConfigServiceClient()` on the created embedded KV store.


robskillington:
@cw9 this test `TestMultipleWatchesFromNotExist` keeps flaking, can we fix this?  Landing to keep making progress.

PR #185, opened 2017-10-20
Fix nil logger specified by default with kv.Configuration

robskillington:


cw9:
Hey on my way to a soccer game now but this means any config that has no logger defined will use nillogger, which felt wrong, can you check with Xi for his logger use case on this? Probably it can be checked by the csclient and assign valid logger there

PR #187, opened 2017-10-21
Use logger in config service client as default logger for kv client

cw9:
Seems current logic means the kv client in m3db will be using nulllogger as well, we probably need to upgrade all repos with this change, seems the only place with a real logger with the previous version is the one in m3aggregator where we explicitly set a logger with Warn level.

@robskillington @xichen2020 @prateek @jeromefroe

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"sync"
 
@@ -111,8 +112,9 @@ func (c *csclient) Txn() (kv.TxnStore, error) {
 	c.txnOnce.Do(func() {
 		kvOpts := kv.NewOptions().
 			SetNamespace(kvPrefix).
-			SetEnvironment(c.opts.Env())
-		c.txn, c.txnErr = c.createTxnStore(kvOpts)
+			SetEnvironment(c.opts.Env()).
+			SetLogger(c.logger)
+		c.txn, c.txnErr = c.TxnStore(kvOpts)
 	})
 	return c.txn, c.txnErr
 }
@@ -311,7 +313,7 @@ func validateTopLevelNamespace(namespace string) error {
 }
 
 func (c *csclient) sanitizeOptions(opts kv.Options) (kv.Options, error) {
-	if opts.Logger() == nil {
+	if opts.Logger() == nil || reflect.DeepEqual(opts.Logger(), log.NullLogger) {
 		opts = opts.SetLogger(c.logger)
 	}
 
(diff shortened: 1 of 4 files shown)


client/etcd/client.go
```diff
@@ -311,7 +309,7 @@ func validateTopLevelNamespace(namespace string) error {
 }
 
 func (c *csclient) sanitizeOptions(opts kv.Options) (kv.Options, error) {
-	if opts.Logger() == nil {
+	if opts.Logger() == nil || reflect.DeepEqual(opts.Logger(), log.NullLogger) {
```
xichen2020: We can probably just do `opts.Logger() == log.NullLogger` here?

cw9: We can't the NullLogger is not comparable, I guess because of the slice of fields in it.

xichen2020: Sure, but `log.NullLogger` is a singleton and of type `Logger`, which is an interface, we should be able to compare interface pointers no?

cw9: Are you saying we should compare the address of the struct under the log.Logger interface? but I don't see how can we we get that without using reflect or unsafe pointer? Please feel free to commandeer this pr if there is an easy way to do it

xichen2020: What I meant was `NullLogger` is a singleton and of `Logger` type, which is an interface. Therefore under the hood, `NullLogger` contains a pointer to the slice fields in the null logger on the heap. As such, checking `opts.Logger() == log.NullLogger` should just work. FWIW I just tried it and it builds locally

xichen2020: Ah I see, didn't realize the `NullLogger` wasn't a pointer type by itself, which would cause a panic during runtime. Put up a m3x PR here to make `NullLogger` a pointer type: https://github.com/m3db/m3x/pull/108 and updated this PR to use direct logger comparison without reflection.

client/etcd/client.go
```diff
@@ -109,10 +109,7 @@ func (c *csclient) KV() (kv.Store, error) {
 
 func (c *csclient) Txn() (kv.TxnStore, error) {
 	c.txnOnce.Do(func() {
-		kvOpts := kv.NewOptions().
-			SetNamespace(kvPrefix).
-			SetEnvironment(c.opts.Env())
-		c.txn, c.txnErr = c.createTxnStore(kvOpts)
+		c.txn, c.txnErr = c.TxnStore(kv.NewOptions())
```
jeromefroe: We don't need to set the namespace and environment here anymore?

xichen2020: That's taken care of in `sanitizeOptions`.

jeromefroe: Oh, I see, Github had collapsed those lines in my view lol.

PR #191, opened 2017-11-20
[leader] Expand unit test for dead leader session

schallert:
We want to ensure that even if a session dies we can still look up the
leader. This exercises that behavior in the unit tests.


services/leader/election/client_test.go
```diff
@@ -145,6 +145,13 @@ func TestCampaign_DeadSession(t *testing.T) {
 	case <-time.After(30 * time.Second):
 		t.Error("should receive err from client2 after orphaning session")
 	}
+
+	// expect that even though cl2 has a dead session it can still lookup leader
```
xichen2020: nit: `Expect...`.

PR #192, opened 2017-11-26
Upgrade etcd client to 3.2.10 to pick up latest client-side load balancer fixes

xichen2020:
cc @cw9 @schallert 

This PR upgrades etcd client to 3.2.10 and grpc to 1.7.3, along with related dependency upgrades. It appears that the upgrade has fixed an issue seen with etcd before where a small percentage of clients do not receive key-value updates.

This upgrade however does require go version >= 1.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 go:
-  - 1.7.6
-  - 1.8.3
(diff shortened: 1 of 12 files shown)


schallert:
LGTM

PR #196, opened 2017-12-04
Reuse the same shard set id when adding back leaving instances

cw9:
Also included a change to start the ShardSetID from 1 instead of 0.
@xichen2020

### DIFF
--- a/placement/selector/mirrored.go
+++ b/placement/selector/mirrored.go
@@ -357,18 +357,11 @@ func assignShardsToGroupedInstances(
 	var (
 		instances   = make([]placement.Instance, 0, p.ReplicaFactor()*len(groups))
 		shardSetIDs = nextNShardSetIDs(p, len(groups))
+		ssID        uint32
 	)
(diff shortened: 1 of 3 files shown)


placement/selector/mirrored.go
```diff
@@ -358,10 +358,26 @@ func assignShardsToGroupedInstances(
...
-		ssID := shardSetIDs[i]
+	for _, group := range groups {
+		var (
+			useNewSSID bool
+			ssID       uint32
+		)
+		for _, instance := range group {
+			_, exist := p.Instance(instance.ID())
+			if !exist {
+				useNewSSID = true
```
xichen2020: Do we want to validate that all of them (instead of only a subset of them) are in the group (or not in the group)?

cw9: I think we should, actually I think the check should be more strict so that all the instances in the same group needs to have the same shard set id to avoid be assigned a new one.

PR #197, opened 2017-12-12
Expose history function in placement storage interface

cw9:
So we can read placement in the history
@prateek @xichen2020

### DIFF
--- a/placement/storage/helper.go
+++ b/placement/storage/helper.go
@@ -52,8 +52,8 @@ type Helper interface {
 	// ValidateProto validates if the given proto message is valid for placement.
 	ValidateProto(proto proto.Message) error
 
-	// History returns the placement of a specific version.
-	History(version int) (placement.Placement, error)
+	// PlacementForVersion returns the placement of a specific version.
+	PlacementForVersion(version int) (placement.Placement, error)
 }
 
 // newHelper returns a new placement storage helper.
@@ -77,7 +77,7 @@ func newPlacementHelper(store kv.Store, key string) Helper {
 	}
 }
 
-func (h *placementHelper) History(version int) (placement.Placement, error) {
+func (h *placementHelper) PlacementForVersion(version int) (placement.Placement, error) {
 	values, err := h.store.History(h.key, version, version+1)
 	if err != nil {
 		return nil, err
@@ -206,7 +206,7 @@ func (h *stagedPlacementHelper) placements() (placement.Placements, int, error) 
 	return ps, value.Version(), err
 }
 
-func (h *stagedPlacementHelper) History(version int) (placement.Placement, error) {
+func (h *stagedPlacementHelper) PlacementForVersion(version int) (placement.Placement, error) {
 	values, err := h.store.History(h.key, version, version+1)
 	if err != nil {
 		return nil, err
--- a/placement/storage/helper.go
+++ b/placement/storage/helper.go
@@ -84,7 +84,7 @@ func (h *placementHelper) PlacementForVersion(version int) (placement.Placement,
 	}
 
 	if len(values) != 1 {
-		return nil, fmt.Errorf("invalid number of values returned: %d, expecting 1", len(values))
+		return nil, fmt.Errorf("invalid number of placements returned: %d, expecting 1", len(values))
 	}
 
 	return placementFromValue(values[0])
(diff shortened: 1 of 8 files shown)


xichen2020:
Hm why do we need to export this? I'm actually in favor of removing the `History` API altogether since it's semantics is not well defined and in general seems like an overly strong requirement for a general key value store

prateek:
Chao put this up because of something I ran into this afternoon. A join node in M3DB triggered a panic. I was trying to look at past placements to see shard distribution for a theory and we figured it'd be useful to have an API to enable this in M3Cluster directly. 

Similar to this case, I'm sure people would like the ability to look at a bounded history of changes per key. Lack of it would make debugging/auditing changes made to KV much harder, if not impossible. 

Would defining explicit semantics (e.g. bounding to X historical entries per key) help allay the concern?

placement/types.go
```diff
@@ -431,6 +431,9 @@ type Storage interface {
 
 	// Proto returns the placement proto.
 	Proto() (proto.Message, int, error)
+
+	// History returns the placement of a specific version.
+	History(version int) (Placement, error)
```
prateek: Do we need a way to get the past version numbers themselves? Or are versions always incremented?

xichen2020:
Thanks for the usecase, that helps to provide some context for the change.

Just to elaborate a little, I have two main concerns:
* The `History` API as part of the `Store` interface puts too strong a requirement on the underlying implementation. It requires that all implementations of this interface keep track of multiple versions (otherwise the current API should do) or return a NotImplemented error, which doesn't feel very clean to me. For example, a Cassandra based implementation that only keeps the last version using `CheckAndSet` would need to return an error here. If there is really such a need, I'd suggest moving the `History` and its relatives to a separate `MultiVersionedStore` interface and pull that out of the generic `Store` interface. This way when we move the KV related interface and implementation out of m3cluster into a standalone repo just for KV stuff, it'll make a lot more sense and requires less refactoring.

* The semantics for `History` is not well defined. Specifically, each implementation of the KV store may have their own way of retaining versions, if at all. For example, an implementation that only keeps the last version can only return a valid result for the last version. Another implementation that compacts data aggressively (e.g., etcd) may return different results depending on the compaction settings and when the `History` API is called. Yet another implementation that retains all histories indefinitely will always return valid results. This means different implementations of the KV store will have different semantics for the `History` API, which IMO puts a lot of cognitive burden on the users of this API since they essentially need to know the peculiarities of the different implementations since the semantics is inconsistent across implementations.

placement/types.go
cw9: versions on the placement are always incremented, when you get the latest placement you will see the current version, then you can refer the previous versions you want to look up

cw9:
I agree that the history API could be pulled into MultiVersionedStore for better isolation, but I don't understand the concern about different implementations, if the store contains the version(s) requested, it returns the value(s) for the version(s), if the store does not contain the version(s) requested, for example the versions(s) could be even be invalid, or the history has been compacted during specific implementation, then the store should return errors. As long as we define a clear semantic on the interface then different implementations would just follow the semantic, no?

xichen2020:
You could define the semantic of the `History` API that way yes, that will be consistent across implementations, but IMO the semantic is a bit unconventional. That said, I'd be less concerned if it's pulled into its own interface.

cw9:
Sure, the original History API was mainly crafted for R2 usage, we could revisit the semantic there , I feel changing the interface from `History(from, to)` to `GetVersion(version)` is probably cleaner, but that should be a seperate diff. 

That said, I'm also going to rename the new function in placement storage from `History()` to `PlacementForVersion` to decouple the function names

placement/storage/helper.go
```diff
@@ -74,6 +77,20 @@ func newPlacementHelper(store kv.Store, key string) Helper {
...
 }
 
+func (h *placementHelper) PlacementForVersion(version int) (placement.Placement, error) {
+	values, err := h.store.History(h.key, version, version+1)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(values) != 1 {
+		return nil, fmt.Errorf("invalid number of values returned: %d, expecting 1", len(values))
```
xichen2020: `invalid number of placements` might be more clear.

placement/storage/helper.go
```diff
@@ -74,6 +77,20 @@ func newPlacementHelper(store kv.Store, key string) Helper {
...
+	}
+
+	p, err := placementFromValue(values[0])
```
xichen2020: nit: `return placementFromValue(values[0])`

PR #199, opened 2018-01-24
Update some internal interfaces

cw9:
Hide some functions/interfaces that are not used outside the package and can change internal state

@xichen2020

### DIFF
--- a/placement/algo/sharded_helper.go
+++ b/placement/algo/sharded_helper.go
@@ -59,16 +59,8 @@ const (
 
 type assignLoadFn func(instance placement.Instance) error
 
-// PlacementHelper helps the algorithm to place shards.
-type PlacementHelper interface {
-	// Instances returns the list of instances managed by the PlacementHelper.
-	Instances() []placement.Instance
-
-	// HasRackConflict checks if the rack constraint is violated when moving the shard to the target rack.
-	HasRackConflict(shard uint32, from placement.Instance, toRack string) bool
-
-	// placeShards distributes shards to the instances in the helper, with aware of where are the shards coming from.
-	placeShards(shards []shard.Shard, from placement.Instance, candidates []placement.Instance) error
+type placementHelper interface {
+	PlacementHelper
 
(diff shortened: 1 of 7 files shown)


placement/algo/sharded_helper.go
```diff
@@ -67,25 +67,25 @@ type PlacementHelper interface {
 	// HasRackConflict checks if the rack constraint is violated when moving the shard to the target rack.
 	HasRackConflict(shard uint32, from placement.Instance, toRack string) bool
 
-	// PlaceShards distributes shards to the instances in the helper, with aware of where are the shards coming from.
-	PlaceShards(shards []shard.Shard, from placement.Instance, candidates []placement.Instance) error
+	// placeShards distributes shards to the instances in the helper, with aware of where are the shards coming from.
```
xichen2020: It's somewhat unconventional to have public interfaces with private APIs. If we don't want users to use these APIs, perhaps we should just remove them or move them to a private interface?

cw9: Sounds good, separated the interfaces

placement/algo/sharded_helper.go
```diff
@@ -67,28 +88,11 @@ type PlacementHelper interface {
...
-
-	// ReclaimLeavingShards reclaims all the leaving shards on the given instance
-	// by pulling them back from the rest of the cluster.
-	ReclaimLeavingShards(instance placement.Instance)
-
-	// ReturnInitializingShards returns all the initializing shards on the given instance
-	// by returning them back to the original owners.
-	ReturnInitializingShards(instance placement.Instance)
+	// placeShards distributes shards to the instances in the helper, with aware of where are the shards coming from.
+	placeShards(shards []shard.Shard, from placement.Instance, candidates []placement.Instance) error
```
xichen2020: Should this be moved as well?

cw9: lol, missed this. Good catch

PR #200, opened 2018-02-05
Allow replace with less weight in instance selector

cw9:
Allow replace with less weight when partial replace is allowed by the options. The algorithm already supports that, in which case the algorithm will fill up the new host and distribute the rest of the load to the rest of the cluster evenly. But right now instance selector is returning errors regardless of the options.

This only affects the non-mirrored mode, while mirrored mode does not allow partial replace and requires the weight to be equal between the new and old instances.

@xichen2020

### DIFF
--- a/placement/selector/non_mirrored.go
+++ b/placement/selector/non_mirrored.go
@@ -34,22 +34,22 @@ var (
 	errNoValidInstance = errors.New("no valid instance in the candidate list")
 )
 
-type nonMirroredFilter struct {
+type nonMirroredSelector struct {
 	opts placement.Options
 }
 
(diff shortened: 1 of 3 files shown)


placement/selector/non_mirrored.go
```diff
@@ -144,7 +144,7 @@ func (f *nonMirroredFilter) SelectReplaceInstances(
 	}
 	result, leftWeight := fillWeight(groups, int(totalWeight))
 
-	if leftWeight > 0 {
+	if leftWeight > 0 && !f.opts.AllowPartialReplace() {
```
cw9: Here is the only meaningful change besides renaming and test change

PR #203, opened 2018-03-06
Add a comment for kv store

cw9:


kv/etcd/store.go
```diff
@@ -395,17 +395,15 @@ func (c *client) update(key string, events []*clientv3.Event) error {
 		nv  kv.Value
 		err error
 	)
-
 	if len(events) == 0 {
-		nv, err = c.getFromKVStore(key)
+		if nv, err = c.getFromKVStore(key); err != nil {
```
xichen2020: nit: move err def here.

cw9: sounds good

PR #204, opened 2018-03-07
Use gogo/protobuf

justinjc:
* Update .ci to use gogo/protobuf generation
* Regenerate protos


cw9:
The change looks fine to me, could you please update statsdex/m3admin to use this version and make sure all the kv get/set + placement get/set still works as is? Make sure to use a test/dummy environment during the test

justinjc:
Okay sure.

PR #205, opened 2018-03-07
Add an option to control the shard states in the placement

cw9:
1, Added ShardStateType in placement/types.go: 
- By default we will use TransitionShardState, which includes Initializing, Available, Leaving states, for M3DB and M3Agg
- SimpleShardState will only have shards in Available state, can be used by the cache or indexer

2, Algo will pick up the new option in tryCleanupShardState()

@xichen2020 @robskillington

### DIFF
--- a/placement/algo/mirrored.go
+++ b/placement/algo/mirrored.go
@@ -74,17 +74,7 @@ func (a mirroredAlgorithm) InitialPlacement(
 		return nil, err
 	}
 
-	p, err := placementFromMirror(mirrorPlacement, instances, rf)
-	if err != nil {
-		return nil, err
-	}
-
-	// NB(cw): Do not validate shards for initial placement.
-	p, _, err = markAllShardsAvailable(
-		p,
-		a.opts.SetIsShardCutoverFn(nil).SetIsShardCutoffFn(nil),
-	)
-	return p, err
+	return placementFromMirror(mirrorPlacement, instances, rf)
 }
 
 func (a mirroredAlgorithm) AddReplica(p placement.Placement) (placement.Placement, error) {
--- a/placement/algo/sharded.go
+++ b/placement/algo/sharded.go
@@ -119,7 +119,6 @@ func (a rackAwarePlacementAlgorithm) RemoveInstances(
 			return nil, err
 		}
 	}
-
 	return tryCleanupShardState(p, a.opts)
 }
 
--- a/placement/algo/sharded.go
+++ b/placement/algo/sharded.go
@@ -58,25 +58,20 @@ func (a rackAwarePlacementAlgorithm) InitialPlacement(
 		return nil, err
 	}
 
-	p := ph.generatePlacement()
+	var (
+		p   = ph.generatePlacement()
+		err error
+	)
 	for i := 1; i < rf; i++ {
-		ph := newAddReplicaHelper(p, a.opts)
-		if err := ph.placeShards(newShards(p.Shards()), nil, ph.Instances()); err != nil {
+		p, err = a.AddReplica(p)
+		if err != nil {
 			return nil, err
 		}
-		if err := ph.optimize(safe); err != nil {
-			return nil, err
-		}
-		p = ph.generatePlacement()
 	}
 
 	// NB(r): All new placements should appear as available for
 	// proper client semantics when calculating consistency results
-	p, _, err := markAllShardsAvailable(
-		p,
-		a.opts.SetIsShardCutoverFn(nil).SetIsShardCutoffFn(nil),
-	)
-	return p, err
+	return cleanupShardState(p, a.opts)
 }
 
 func (a rackAwarePlacementAlgorithm) AddReplica(p placement.Placement) (placement.Placement, error) {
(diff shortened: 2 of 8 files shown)


placement/algo/mirrored.go
```diff
@@ -74,17 +74,7 @@ func (a mirroredAlgorithm) InitialPlacement(
 		return nil, err
 	}
 
-	p, err := placementFromMirror(mirrorPlacement, instances, rf)
-	if err != nil {
-		return nil, err
-	}
-
-	// NB(cw): Do not validate shards for initial placement.
```
xichen2020: Do we no longer need this? I'm also not sure why we need this in the first place.. Shouldn't all shards always be marked as available in the initial placement?

placement/algo/sharded.go
```diff
@@ -58,25 +58,20 @@ func (a rackAwarePlacementAlgorithm) InitialPlacement(
...
+		p   = ph.generatePlacement()
+		err error
+	)
 	for i := 1; i < rf; i++ {
-		ph := newAddReplicaHelper(p, a.opts)
-		if err := ph.placeShards(newShards(p.Shards()), nil, ph.Instances()); err != nil {
-			return nil, err
-		}
-		if err := ph.optimize(safe); err != nil {
+		p, err = a.AddReplica(p)
```
xichen2020: Seems we used to only place shards to non leaving instances, I guess that doesn't matter for initial placement?

placement/algo/sharded_helper.go
```diff
@@ -818,6 +818,28 @@ func markShardAvailable(p placement.Placement, instanceID string, shardID uint32
 	return p, nil
 }
 
+func tryCleanupShardState(
+	p placement.Placement,
+	opts placement.Options,
+) (placement.Placement, error) {
+	if opts.ShardStateType() == placement.SimpleShardState {
+		p, err := cleanupShardState(p, opts)
```
xichen2020: nit: `return cleanupShardState(...)`

placement/algo/sharded_helper.go
```diff
@@ -818,6 +818,28 @@ func markShardAvailable(p placement.Placement, instanceID string, shardID uint32
 	return p, nil
 }
 
+func tryCleanupShardState(
+	p placement.Placement,
+	opts placement.Options,
+) (placement.Placement, error) {
+	if opts.ShardStateType() == placement.SimpleShardState {
```
xichen2020: Why is it a problem to mark shards as available for transitioning shards? Thought that was what the old logic does..

placement/algo/mirrored.go
cw9: We no longer need this as the sharded algo is already marking them as available

placement/algo/sharded_helper.go
xichen2020: Can you elaborate on "hard to make MarkShardAsAvailable" calls?

placement/algo/mirrored.go
xichen2020: So it's not needed before?

placement/algo/sharded_helper.go
xichen2020: Looked at this again, so basically you are using the current placement algorithm to generate placements, but for simple shard states, you'll immediately mark them as available (i.e., removing leaving shards and marking initializing shards as available) before returning them to the caller. Is that correct?

If so, perhaps change the option to a boolean `AllowTransitionalStates` option instead of defining `ShardStateType`, and will mark all states as available if the option is false? That IMO feels more intuitive. Right now the `SimpleShardState` is still using `Available` state, so technically it's not a different shard state and that's a bit confusing.

xichen2020: Sure, would it be better to call it `UseSingleShardState`?

cw9: I'd rather say NoShardState or IgnoreShardState, cause it does not matter to user if there is one shard state or more, user should not check the shard state at all, we may also use Available or other states in implementation.

xichen2020: Yeah however we are returning a placement to the user who can technically access the shards and see the shard states as available, so unless we can literally set the shard state to `None`, I think calling it `NoShardState` would be inaccurate. I'm fine with `IgnoreShardStates` though.

placement/types.go
```diff
@@ -414,12 +414,11 @@ type Options interface {
 type ShardStateType int
```
xichen2020: Perhaps we should also rename the type to something else? Feels like this is describing how the placement should be generated rather than the type of shard states. Do you think `ShardStateMode` or `PlacementMode` or `ShardStateConstraint` would be more fitting?

cw9: I think PlacementMode sounds better, also we can do StableShardStateOnly and AllowTransitionalShardStates for the enums

cw9: Actually I'm more leaning towards ShardStateMode now as this mainly controls shard states

placement/algo/sharded_helper.go
```diff
@@ -818,6 +818,27 @@ func markShardAvailable(p placement.Placement, instanceID string, shardID uint32
 	return p, nil
 }
 
+func tryCleanupShardState(
```
robskillington: nit: Maybe a better would be `cleanupShardStateIfRequired(...)`?

PR #206, opened 2018-03-07
Fix flaky test in heartbeat store

cw9:
kv watch by default will do a get() upon creation, which will call heartbeatStore.Get(), for a service with no heartbeats it will return an empty string, so all watch should expect a notification soon after creation regardless whether there is any heartbeat for the service.

@schallert


schallert:
LGTM

PR #207, opened 2018-03-09
Add type safe watch for service watches

robskillington:


services/types.go
```diff
@@ -86,6 +87,34 @@ type Services interface {
...
+	// C returns the notification channel
+	C() <-chan struct{}
+
+	// Get returns the latest service of the service watchable
+	Get() Service
+}
+
+// Watchable is a container of a service.Service value
+// that can be subscribed to by issuing a watch
+type Watchable interface {
```
cw9: Can skip the Service specific watchable and only expose ServiceWatch? I think internally using a xwatch.Watch should be safe enough?

PR #210, opened 2018-03-11
Remove the LooseRackCheck option

cw9:
This is a prerequisite diff for renaming rack to isolation group. The option is very legacy and unsafe, we use to need this because M3DB could not do add/remove and can only do replace, so we had this option just in case we need to replace host even with rack conflict, but that is kinda dangerous and now we always can do add/remove in all the services that use m3cluster we should remove this.

@xichen2020 @robskillington


services/types.go
```diff
@@ -89,7 +88,7 @@ type Services interface {
 
 // Watch is a watcher that issues notification when a service is updated
 type Watch interface {
-	xclose.SimpleCloser
+	Close()
```
xichen2020: nit: Add comments here.

xichen2020:
Can you elaborate a little on why this is preventing us from renaming rack to isolation group?

cw9:
Not really blocking us, just I can save the work to rename this option to `LooseIsolationGroupCheck`

xichen2020:
Understood this wasn't so much of a problem for us, but do you foresee people in the OSS world will need this option in the future?

cw9:
I think when people really need to violate rack constraint, it's probably more explicit to override rack to something unique explicitly, for example: "sjc1-foo-temp", to be clear that this instance should be fixed soon, rather than just leaving rack conflicted shards in the placement and later hard to recall why they are there.

xichen2020:
Ideally yes, however there are occasions where that's not reasonable or possible, e.g., if the users of such library has no permission to change the naming of such racks. They could also do what you suggested by renaming the rack for conflicting shards in their code to something else to place conflicting shards on the same rack, but that to me feels like a hack that's rather difficult to keep track of and clean up.

xichen2020:
Btw, if you feel this is something that people would almost never want and pollutes the codebase with additional code complexity, and is something that can be added back if so needed, then I agree and am fine with that.

xichen2020:
I agree leaving conflicted shards in the placement is undesirable, but sometimes you simply don't have a choice (e.g., if the only server you have is on the same rack as other servers), and I don't see how renaming the rack to something else to work around the restriction of not allowing conflicting shards on the same rack is making it any easier (if anything, it'll actually make it harder to clean up since now you have to clean up both the conflicting shards and the renamed rack).

I do agree however this should be rarely needed, although I think the overhead of leaving it there is not significantly based on the size of this diff. If you feel that makes things difficult to maintain, I'm fine w/ that as long as we can add it back if the need to do so arises.

cw9:
I don't see why changing the rack of the new instances would make it harder to clean up than leaving the conflicted shards in the placement, right now there is nothing that can mark a shard is actually violating rack conflict in the placement, once you run the placement with this option the placement could look perfectly normal even though there are rack conflicted shards in it. But when you have to use instance with rack conflict during a replace and you changed the rack of the instance you are adding, you can actually use that as a note to mark its violation.
Also when the user later have other instances that can be used to fix the situation, they can replace the instance with rack conflict easily with no further rack overriding.

xichen2020:
Sure, but you need to make code changes to actually rename the rack for the conflicting shards and it'll probably look pretty ugly (e.g., if the server is xxx, then its rack is yyy, etc), and then remember that you have made such changes and map it back to the placement, and then clean up that code change if you pushed it to prod, as opposed to just passing in a command line flag.

cw9:
I don't think any mapping is needed, you don't need to change any existing racks, but just override the rack value of the new instance, (which can be done as an extra parameter in the m3admin for example), even if instance from same rack already exist in the placement, you don't need to touch them at all. You also don't need to clean that up, the next time when you use an instance from a new rack, you can just replace the bad instance out with the good instance with no overriding or anything. I think the overriding process at least give you a chance to make that note that instance was "bad" in the placement, rather than the shards where you could potentially lose track of and forgot why they are violating the rack check.

PR #211, opened 2018-03-12
Rename rack to isolation group

cw9:
@xichen2020 @robskillington


placement/algo/sharded_helper.go
```diff
@@ -88,15 +88,15 @@ type PlacementHelper interface {
 	// Instances returns the list of instances managed by the PlacementHelper.
 	Instances() []placement.Instance
 
-	// HasRackConflict checks if the rack constraint is violated when moving the shard to the target rack.
-	HasRackConflict(shard uint32, from placement.Instance, toRack string) bool
+	// CanMove checks if the shard can be moved from the instance to the target isolation group.
+	CanMove(shard uint32, from placement.Instance, isolationGroup string) bool
```
robskillington: Isn't this method the inverse of `CanMove`?  It seems to only return true if you cannot move it (i.e. there is a conflict).

Would a better name be `CannotPlaceShardInIsolationGroupFromInstance(...)`?  Also using parameters `...(shard, isolationGroup, placement.Instance)` might make more sense to match the order of things described by the name.

placement/selector/mirrored.go
```diff
@@ -419,42 +419,42 @@ func (h host) addInstance(port uint32, instance placement.Instance) error {
...
+		return fmt.Errorf("could not add instance %s to host %s, isolation group mismatch: %s and %s",
+			instance.ID(), h.name, instance.IsolationGroup(), h.ig)
 	}
 	h.portToInstance[port] = instance
 	return nil
 }
 
-type rack struct {
-	rack  string
+type ig struct {
```
robskillington: Yeah we should probably call these `type isoGroup struct` or something and put a comment above it.

placement/selector/mirrored.go
```diff
@@ -419,42 +419,42 @@ func (h host) addInstance(port uint32, instance placement.Instance) error {
...
+	ig    string
 	hosts []host
 }
 
-type racksByNumHost []*rack
+type igsByNumHost []*ig
```
robskillington: Similar here `type isoGroupsByNumHost []*isoGroup`?

placement/algo/sharded_helper.go
cw9: Yeah I think making it `CanMoveShard` is probably more clear, I'm going to reverse the implementation and usage instead

generated/proto/placementpb/placement.proto
```diff
@@ -41,7 +41,7 @@ message Placement {
 
 message Instance {
   string id = 1;
-  string rack = 2;
+  string isolation_group = 2;
```
xichen2020: I think this is backward compatible, but might be worth double checking.

placement/algo/sharded_helper.go
```diff
@@ -88,15 +88,15 @@ type PlacementHelper interface {
...
+	// CanMoveShard checks if the shard can be moved from the instance to the target isolation group.
+	CanMoveShard(shard uint32, fromInstance placement.Instance, toIsolationGroup string) bool
 }
 
 type helper struct {
 	targetLoad         map[string]int
 	shardToInstanceMap map[uint32]map[placement.Instance]struct{}
-	rackToInstancesMap map[string]map[placement.Instance]struct{}
-	rackToWeightMap    map[string]uint32
+	igToInstancesMap   map[string]map[placement.Instance]struct{}
```
xichen2020: nit: would it be more clear to call it `groupToInstancesMap`? `ig` is a bit vague IMO.

placement/algo/sharded_helper.go
```diff
@@ -241,11 +241,11 @@ func (ph *helper) scanCurrentLoad() {
 }
 
 func (ph *helper) buildTargetLoad() {
-	overWeightedRack := 0
+	overWeightedIG := 0
```
xichen2020: nit: I'd spell out IG here.

placement/service/service_test.go
```diff
@@ -271,7 +272,7 @@ func TestBadAddReplica(t *testing.T) {
 		10, 1)
 	assert.NoError(t, err)
 
-	// not enough racks/instances
+	// not enough isolation groups/instances
```
xichen2020: nit: comment should start w/ Upper case letter.

placement/service/service_test.go
```diff
@@ -320,7 +321,7 @@ func TestBadRemoveInstance(t *testing.T) {
 	_, err = p.RemoveInstances([]string{"not_exist"})
 	assert.Error(t, err)
 
-	// not enough racks/instances after removal
+	// not enough isolation groups/instances after removal
```
xichen2020: nit: comment should start w/ Upper case letter.

placement/service/service_test.go
```diff
@@ -353,7 +354,7 @@ func TestBadReplaceInstance(t *testing.T) {
 	)
 	assert.Error(t, err)
 
-	// not enough rack after replace
+	// not enough isolation groups after replace
```
xichen2020: nit: comment should start w/ Upper case letter.

generated/proto/placementpb/placement.proto
cw9: Yeah verified this is backward compatible

PR #212, opened 2018-03-13
Update mem value to behave the same as kv value

cw9:
Previously when you delete a key from mem store and set a new value for the key, the new value will be of version 1 which might not look like newer than the value before the deletion, this pr adds a revision field similar to what kv store has.

@xichen2020

### DIFF
--- a/kv/mem/store.go
+++ b/kv/mem/store.go
@@ -141,15 +141,7 @@ func (s *store) setWithLock(key string, val proto.Message) (int, error) {
 		lastVersion = vals[len(vals)-1].version
 	}
 	newVersion := lastVersion + 1
-	s.revision++
-	fv := &value{
-		version:  newVersion,
-		revision: s.revision,
-		data:     data,
-	}
-	s.values[key] = append(vals, fv)
-	s.updateWatchable(key, fv)
-
+	s.updateInternal(key, newVersion, data)
 	return newVersion, nil
 }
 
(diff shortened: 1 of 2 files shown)


kv/mem/store.go
```diff
@@ -56,16 +56,28 @@ func NewValueWithData(vers int, data []byte) kv.Value {
...
 
 func (v value) Version() int                      { return v.version }
 func (v value) Unmarshal(msg proto.Message) error { return proto.Unmarshal(v.data, msg) }
-func (v value) IsNewer(other kv.Value) bool       { return v.version > other.Version() }
+func (v value) IsNewer(other kv.Value) bool {
+	otherValue, ok := other.(*value)
+	if !ok {
+		return v.version > other.Version()
+	}
+	if v.revision == otherValue.revision {
```
xichen2020: Should we just do `return v.revision > otherValue.revision` instead?

kv/mem/store.go
```diff
@@ -183,14 +182,19 @@ func (s *store) CheckAndSet(key string, version int, val proto.Message) (int, er
 	}
 
 	newVersion := version + 1
+	s.updateInternal(key, newVersion, data)
+	return newVersion, nil
+}
+
+func (s *store) updateInternal(key string, newVersion int, data []byte) {
```
xichen2020: nit: `updateInternalWithLock` might be better, it took me a few seconds to figure out this is called from within a `XXXWithLock` function.

cw9: sg

kv/mem/store.go
cw9: nope, there are lots of places where mem.Value was just created with a value but no revision, when revision is 0 it should just compare version. External users should also not care about the revision as well which is why there is why this is not accessible  externally

xichen2020: Hm it seems if the values are created by reading from the mem stores they should always have revision set. Are you saying people directly create `mem.Value` and mess it? But how do they update the versions in that case?

cw9: The constructor of the mem.Value allows users to set version, people have been creating mem.Value in many tests

PR #213, opened 2018-03-14
Move services impl out of etcd folder as it only depends on kv.Store

cw9:
Had to rename the original services.Options to services.OverrideOptions and rename etcd.Options to services.Options.
Not much real code change beside moving code out of the etcd folder.
@xichen2020 @robskillington @schallert


services/services.go
```diff
@@ -23,28 +23,611 @@ package services
...
+	xwatch "github.com/m3db/m3x/watch"
 
-const (
-	defaultLeaderTimeout = 10 * time.Second
-	defaultResignTimeout = 10 * time.Second
+	"github.com/uber-go/tally"
 )
 
 var (
+	defaultGaugeInterval = 10 * time.Second
```
schallert: Can this be a constant?

cw9: yeah, will make it a const

PR #214, opened 2018-03-15
Refactor logging in kv.Store

cw9:
Remove verbose logging, delegate some info logging to users
Rename kv.Options to kv.OverrideOptions
Remove logger in overrideOptions
Cache kv.Store in the client


kv/util/runtime/value.go
```diff
@@ -168,8 +168,10 @@ func (v *value) updateWithLock(update kv.Value) error {
 		return errNilValue
 	}
 	if v.currValue != nil && !update.IsNewer(v.currValue) {
+		v.log.Warnf("ignore kv update with version %d which is older than the version of the current value: %d", update.Version(), v.currValue.Version())
 		return nil
 	}
+	v.log.Infof("received kv update with version %d for key %s", v.key, update.Version())
```
xichen2020: Swap key and version

kv/util/runtime/value.go
```diff
@@ -168,8 +168,10 @@ func (v *value) updateWithLock(update kv.Value) error {
 		return errNilValue
 	}
 	if v.currValue != nil && !update.IsNewer(v.currValue) {
+		v.log.Warnf("ignore kv update with version %d which is older than the version of the current value: %d", update.Version(), v.currValue.Version())
```
xichen2020: should probably be `which is not newer than ...`

PR #215, opened 2018-03-23
Add placement watch

cw9:
To be used by m3msg

@xichen2020


placement/storage/watch.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3x/log"
+)
+
+type w struct {
+	w      kv.ValueWatch
+	logger log.Logger
+}
+
+func newPlacementWatch(vw kv.ValueWatch, opts placement.Options) placement.Watch {
```
xichen2020: Why not just pass in the logger instead?

placement/storage/watch.go
```diff
@@ -0,0 +1,62 @@
...
+package storage
+
+import (
+	"github.com/m3db/m3cluster/kv"
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3x/log"
+)
+
+type w struct {
+	w      kv.ValueWatch
```
xichen2020: Just embed the value watch?

placement/storage/watch.go
```diff
@@ -0,0 +1,62 @@
...
+func (w *w) Get() placement.Placement {
+	v := w.w.Get()
+	if v == nil {
+		w.logger.Errorf("received nil placement from kv")
+		return nil
+	}
+	p, err := placementFromValue(v)
+	if err != nil {
+		w.logger.Errorf("could not process placement from kv with version %d, %v", v.Version(), err)
+		return nil
```
xichen2020: Hm should we just return err here instead?

placement/storage/watch.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	w.logger.Infof("successfully processed placement from kv with version %d", v.Version())
```
xichen2020: Is this too verbose? I thought we decided against logging value updates in the library?

cw9: Yeah, I don't think this logging will be too verbose, this will be the only log that indicates a successful placement update with the version.

placement/storage/watch.go
cw9: We could, I used the interface without returning error to keep the same convention used by other watches like ValueWatch or ServiceWatch, I could do `Get() (Placement, error)` here but then for the `delete` case I need to return `nil, errors.New("placement is not available")` which is ok but a little weird.

placement/storage/watch.go
cw9: sure

placement/storage/watch.go
cw9: sounds good

placement/storage/watch.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func newPlacementWatch(vw kv.ValueWatch, logger log.Logger) placement.Watch {
+	return &w{
+		ValueWatch: vw,
+		logger:     logger,
+	}
+}
+
+func (w *w) C() <-chan struct{} {
```
xichen2020: FYI You don't need to implement this. The embedded struct already provides it.

placement/storage/watch.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		w.logger.Errorf("could not process placement from kv with version %d, %v", v.Version(), err)
+		return nil, err
+	}
+	w.logger.Infof("successfully processed placement from kv with version %d", v.Version())
+	return p, nil
+}
+
+func (w *w) Close() {
+	w.ValueWatch.Close()
```
xichen2020: FYI You don't need to implement this. The embedded struct already provides it.

placement/storage/watch.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3x/log"
+)
+
+var (
+	errPlacementNotAvailable = errors.New("placement is not available")
+)
+
+type w struct {
+	kv.ValueWatch
```
xichen2020: Blank line after embedded type.

placement/storage/watch.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (w *w) C() <-chan struct{} {
+	return w.ValueWatch.C()
+}
+
+func (w *w) Get() (placement.Placement, error) {
+	v := w.ValueWatch.Get()
+	if v == nil {
+		w.logger.Errorf("received nil placement from kv")
```
xichen2020: If you are returning an error, do you still need to log here as opposed to logging in the caller?

cw9: yeah in that case, I'll move all the logging out

placement/storage/watch.go
```diff
@@ -34,35 +33,20 @@ var (
 
 type w struct {
```
xichen2020: nit: You can just type alias it?

xichen2020:
Also looks like the tests are failing?

placement/storage/watch.go
cw9: I can't as I need to call the ValueWatch.Get() in the Get() function

PR #216, opened 2018-03-23
[selector] Add option to add all candidates in non-mirrored selector

jeromefroe:
This diff adds an option to allow the non-mirrored selector to add all candidates at once. This will be beneficial in caching tiers where each host may run more than one instance of the caching process.

### DIFF
--- a/placement/selector/non_mirrored.go
+++ b/placement/selector/non_mirrored.go
@@ -58,39 +58,15 @@ func (f *nonMirroredSelector) SelectAddingInstances(
 		return nil, err
 	}
 
-	candidateGroups := buildIsolationGroupMap(candidates)
-
-	existingGroups := buildIsolationGroupMap(p.Instances())
-
-	// If there is a isolation group not in the current placement, prefer the isolation group.
-	for r, instances := range candidateGroups {
-		if _, exist := existingGroups[r]; !exist {
-			// All the isolation groups have at least 1 instance.
-			return instances[:1], nil
-		}
+	if f.opts.AddAllCandidates() {
+		return candidates, nil
 	}
 
-	// Otherwise sort the isolation groups in the current placement
-	// by capacity and find a instance from least sized isolation group.
-	groups := make(sortableValues, 0, len(existingGroups))
-	for group, instances := range existingGroups {
-		weight := 0
-		for _, i := range instances {
-			weight += int(i.Weight())
-		}
-		groups = append(groups, sortableValue{value: group, weight: weight})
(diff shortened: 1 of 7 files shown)


placement/selector/non_mirrored.go
```diff
@@ -242,6 +218,45 @@ func buildIsolationGroupMap(candidates []placement.Instance) map[string][]placem
...
 }
 
+func selectSingleCandidate(
+	candidates []placement.Instance,
+	p placement.Placement,
+) (placement.Instance, error) {
+	candidateGroups := buildIsolationGroupMap(candidates)
+	existingGroups := buildIsolationGroupMap(p.Instances())
+
+	// If there is a isolation group not in the current placement, prefer the isolation group.
```
cw9: nit: If there is `an` isolation group

placement/selector/non_mirrored_test.go
```diff
@@ -273,3 +274,66 @@ func TestFilterZones(t *testing.T) {
...
+		SetWeight(1)
+	i4 := placement.NewInstance().
+		SetID("i4").
+		SetIsolationGroup("r2").
+		SetWeight(2)
+
+	tests := []struct {
+		name                  string
+		placement             placement.Placement
+		selector              placement.InstanceSelector
```
cw9: nit: just use placement.Options here

placement/selector/non_mirrored.go
```diff
@@ -58,39 +58,15 @@ func (f *nonMirroredSelector) SelectAddingInstances(
...
-
-	existingGroups := buildIsolationGroupMap(p.Instances())
-
-	// If there is a isolation group not in the current placement, prefer the isolation group.
-	for r, instances := range candidateGroups {
-		if _, exist := existingGroups[r]; !exist {
-			// All the isolation groups have at least 1 instance.
-			return instances[:1], nil
-		}
+	if f.opts.AddAllCandidates() {
```
cw9: could you also add some tests in placement/service/ to make sure this options is doing the right thing? For example when you add 10 instances but 5 of them already exist in the placement, verify that only the other 5 got added in the final result?

cw9:
Actually you might also want to double check that `SelectReplaceInstances` is doing the right thing as well, for example if you want to replace 6 instances from 1 host, but you provided 12 instances from 2 host as candidate instances, you want to use all 6 instances from newHost1 rather than 3 instances from newHost1 and 3 from newHost2. You could reuse some code in the mirrored selector which already handles this logic.

PR #217, opened 2018-03-28
[service] Update replace selector to support adding all instances

jeromefroe:
Propagate the option to add all candidates in calls to replace all instances.

### DIFF
--- a/placement/selector/non_mirrored.go
+++ b/placement/selector/non_mirrored.go
@@ -89,7 +89,7 @@ func (f *nonMirroredSelector) SelectReplaceInstances(
 	}
 
 	if f.opts.AddAllCandidates() {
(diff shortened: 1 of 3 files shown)


placement/selector/non_mirrored.go
```diff
@@ -120,12 +136,31 @@ func (f *nonMirroredSelector) SelectReplaceInstances(
...
+	var leavingWeight int
+	for _, instance := range leavingInstances {
+		leavingWeight += int(instance.Weight())
+	}
+	var candidateWeight int
+	for _, instance := range candidates {
+		candidateWeight += int(instance.Weight())
+	}
+
+	if leavingWeight-candidateWeight > 0 && !f.opts.AllowPartialReplace() {
```
cw9: `leavingWeight-candidateWeight > 0` -> `leavingWeight>candidateWeight`

PR #219, opened 2018-04-13
Make runtime.Value more generic

cw9:


xichen2020:
Perhaps what we could do is to pass in a `watchFn` that creates watches and a `GetFn` that takes a watch as a parameter? e.g.,
```
type watchFn func() watch

type watch interface {
  C() <-chan struct{}
  Close()
}

type getFn func(w watch)
```
?
Then in the `getFn` function you can cast the watch to `m3x.Watch` or `kv.ValueWatch` as needed.

kv/util/runtime/options.go
```diff
@@ -45,17 +44,17 @@ type Options interface {
 	// InitWatchTimeout returns the initial watch timeout.
 	InitWatchTimeout() time.Duration
 
-	// SetKVStore sets the kv store.
-	SetKVStore(value kv.Store) Options
+	// SetUpdatableFn sets the updatable function.
+	SetUpdatableFn(value UpdatableFn) Options
```
xichen2020: `SetNewUpdatableFn` might be better?

kv/util/runtime/value.go
```diff
@@ -37,24 +37,33 @@ var (
...
 
 	// Unwatch stops watching for value updates.
 	Unwatch()
 }
 
-// UnmarshalFn unmarshals a kv value and extracts its payload.
-type UnmarshalFn func(value kv.Value) (interface{}, error)
+// Updatable can be updated.
+// TODO: Move to m3x/watch.
+type Updatable interface {
```
xichen2020: Is `ListenOnlyWatch` a better name?

kv/util/runtime/value.go
```diff
@@ -66,68 +75,63 @@ const (
 type value struct {
 	sync.RWMutex
 
-	key              string
-	store            kv.Store
 	opts             Options
 	log              log.Logger
-	unmarshalFn      UnmarshalFn
+	updatableFn      UpdatableFn
```
xichen2020: `newUpdatableFn` is probably better.

kv/util/runtime/value.go
```diff
@@ -37,24 +37,33 @@ var (
...
+
+	// C returns the notification channel for updates.
+	C() <-chan struct{}
+}
+
+// UpdatableFn returns an updatable.
+type UpdatableFn func() (Updatable, error)
+
+// GetFn returns the latest value.
+type GetFn func(updatable Updatable) (interface{}, error)
```
xichen2020: When will this return an error?

kv/util/runtime/value.go
```diff
@@ -141,47 +145,45 @@ func (v *value) Unwatch() {
...
-	if err != nil {
-		err = fmt.Errorf("error unmarshalling value for version %d: %v", update.Version(), err)
-		return err
-	}
-	if err := v.processFn(latest); err != nil {
-		return err
-	}
-	v.currValue = update
-	return nil
+	return v.processFn(update)
```
xichen2020: Hm can we not keep the existing logic here? Presumably we don't want to process the old/same updates multiple times.

kv/util/runtime/value.go
```diff
@@ -141,47 +145,45 @@ func (v *value) Unwatch() {
...
+}
+
+// IsCreateWatchError returns whether the error is a CreateWatchError.
+func IsCreateWatchError(err error) bool {
```
xichen2020: Is this really needed? it's fairly easy to do a type check/switch outside the package unless the error types are not exported, so I'd just avoid the additional cognitive burden by removing it.

placement/staged_placement_watcher.go
```diff
@@ -48,31 +49,41 @@ type stagedPlacementWatcher struct {
...
+	logger        log.Logger
 
 	state     placementWatcherState
 	proto     *placementpb.PlacementSnapshots
 	placement ActiveStagedPlacement
 }
 
 // NewStagedPlacementWatcher creates a new staged placement watcher.
-func NewStagedPlacementWatcher(opts StagedPlacementWatcherOptions) StagedPlacementWatcher {
+func NewStagedPlacementWatcher(opts StagedPlacementWatcherOptions) (StagedPlacementWatcher, error) {
```
xichen2020: Why would this return an error now?

kv/util/runtime/value.go
```diff
@@ -141,47 +145,45 @@ func (v *value) Unwatch() {
...
+func (v *value) updateWithLock(update interface{}) error {
 	if update == nil {
 		return errNilValue
 	}
-	if v.currValue != nil && !update.IsNewer(v.currValue) {
-		v.log.Warnf("ignore kv update with version %d which is not newer than the version of the current value: %d", update.Version(), v.currValue.Version())
-		return nil
-	}
-	v.log.Infof("received kv update with version %d for key %s", update.Version(), v.key)
-	latest, err := v.unmarshalFn(update)
```
xichen2020: Can we not keep this? Otherwise you'll need to make a lot more code changes than necessary.

placement/staged_placement_watcher.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/m3db/m3cluster/kv"
 	"github.com/m3db/m3cluster/kv/util/runtime"
 	"github.com/m3db/m3x/clock"
+	"github.com/m3db/m3x/log"
```
xichen2020: FYI please remember to update the matchers in `m3metrics/matcher` that are using `kv.Value` as well. They are currently used in collector/query/indexer.

kv/util/runtime/value.go
cw9: The kv watch is already depending that same/old updates, also the interface{} here does not have `.Version()`, if you really want to compare versions outside of what kv.ValueWatch already did for you, you could do that comparison in processFn as well.

kv/util/runtime/value.go
cw9: sure can remove this

kv/util/runtime/value.go
xichen2020: I'm thinking about the case where a watch disconnects and reconnects, would it receive an update with the same version? 

We can have a `VersionedUpdatable` interface that contains the `Version()` API, so if `Updatatable` can be cast to `VersionedUpdatable` then we can apply the same logic here.

cw9: `VersionedUpdatable` also won't help as the `Version()` was not exposed on `ValueWatch`, it's exposed on the `Value` you get from the `ValueWatch`, which is now changed into `interface{}`, the `processFn` implementation will need to cast `interface{}` into for example `Value` for processing, and the version comparison can be done there.

kv/util/runtime/value.go
```diff
@@ -37,24 +38,34 @@ var (
...
+
+	// C returns the notification channel for updates.
+	C() <-chan struct{}
+}
+
+// NewUpdatableFn creates an updatable.
+type NewUpdatableFn func() (Updatable, error)
+
+// GetFn returns the latest value.
+type GetFn func(updatable Updatable) (interface{}, error)
```
xichen2020: Thought we were going to return a `Versionable` here?

kv/util/runtime/value.go
```diff
@@ -37,24 +38,34 @@ var (
...
 
 // ProcessFn processes a value.
+// If the value is versionable, then only newer value will be processed.
 type ProcessFn func(value interface{}) error
```
xichen2020: Should this take a `Versionable` instead?

kv/util/runtime/value.go
```diff
@@ -37,24 +38,34 @@ var (
...
 
-// updateWithLockFn updates a value while holding a lock.
-type updateWithLockFn func(value kv.Value) error
+// processWithLockFn updates a value while holding a lock.
+type processWithLockFn func(value interface{}) error
```
xichen2020: Hm I actually think `updateWithLockFn` is a better name here. Also the param should be a `Versionable` I'd think?

kv/util/runtime/value.go
```diff
@@ -141,49 +148,75 @@ func (v *value) Unwatch() {
...
+func (v *value) isNewerUpdate(update interface{}) bool {
+	if v.currValue == nil {
+		return true
+	}
+	versionedCurrValue, ok := v.currValue.(kv.Versionable)
+	if !ok {
+		return true
+	}
+	versionedUpdate, ok := update.(kv.Versionable)
+	if !ok {
```
xichen2020: This is kinda weird, can we just make the `update` always a `Versionable`?

kv/util/runtime/value.go
```diff
@@ -141,49 +148,75 @@ func (v *value) Unwatch() {
...
+		// Current value is already versioned but the update is not versionable.
+		v.log.Warnf(
+			"ignore update without version which is not newer than the version of the current value: %d",
+			versionedCurrValue.Version(),
+		)
+		return false
+	}
+	if versionedUpdate.IsNewer(versionedCurrValue) {
```
xichen2020: Is this comparison logic backwards? i.e., should it be `if versionedCurrValue.IsNewer(versionedUpdate)`? If so, can we add a unit test for this?

placement/staged_placement_watcher.go
```diff
@@ -139,14 +144,23 @@ func (t *stagedPlacementWatcher) toStagedPlacement(value kv.Value) (interface{},
...
-func (t *stagedPlacementWatcher) process(value interface{}) error {
+func (t *stagedPlacementWatcher) process(update interface{}) error {
 	t.Lock()
 	defer t.Unlock()
 
 	if t.state != placementWatcherWatching {
 		return errPlacementWatcherIsNotWatching
 	}
-	ps := value.(StagedPlacement)
+	if update == nil {
```
xichen2020: I think by the time you get here, `update` is guaranteed to be non nil, yeah?

placement/staged_placement_watcher.go
```diff
@@ -139,14 +144,23 @@ func (t *stagedPlacementWatcher) toStagedPlacement(value kv.Value) (interface{},
...
+		return errNilValue
+	}
+	value := update.(kv.Value)
+	t.logger.Infof("processing update from kv for key %s with version %d", t.key, value.Version())
+	ps, err := t.toStagedPlacementWithLock(value)
+	if err != nil {
+		t.logger.Errorf("could not convert kv update to staged placement, %v", err)
```
xichen2020: nit: conventionally should be `... staged placement: %v", err)`

kv/util/runtime/value.go
cw9: Because not all cases can be `Versionable`. For example just `boolean` and `int64` from `watch.Watch`

kv/util/runtime/value.go
cw9: see comment above

kv/util/runtime/value.go
cw9: see comment above

kv/util/runtime/value.go
cw9: yes, and there is a test case for this

placement/staged_placement_watcher.go
cw9: That is depending on the `GetFn` passed in by user, if the user does bad things then here can still be nil, so I'd rather be on the defensive side.

kv/util/runtime/value.go
xichen2020: Not sure if I follow, if you are returning true here, then we shouldn't log the warning message, if you want to return false here, then the condition is backwards.

kv/util/runtime/value.go
xichen2020: Do we want to support a generic `watch.Watch` use case? I'd be happy to restrict it a little and only support cases where the value is always `Versionable` since it leads to much cleaner code. Is this something you need?

kv/util/runtime/value.go
cw9: Yeah that's true, will move the warn logging out

kv/util/runtime/value.go
cw9: I hope that we could leverage this runtime.Value to replace a bunch of Watch util functions in `kvutil` package. But I don't have to use it right now, could start with enforcing `Versionable` to begin with if you feel strongly.

placement/staged_placement_watcher.go
```diff
@@ -59,19 +62,27 @@ type stagedPlacementWatcher struct {
...
 	watcher := &stagedPlacementWatcher{
+		key:           opts.StagedPlacementKey(),
+		logger:        opts.InstrumentOptions().Logger(),
 		nowFn:         opts.ClockOptions().NowFn(),
 		placementOpts: opts.ActiveStagedPlacementOptions(),
 		proto:         &placementpb.PlacementSnapshots{},
 	}
 	watcher.doneFn = watcher.onActiveStagedPlacementDone
 
+	updatableFn := func() (runtime.Updatable, error) {
```
xichen2020: Change `updatableFn` to `newUpdatableFn`.

kv/util/runtime/value.go
```diff
@@ -37,24 +38,34 @@ var (
...
+
+	// C returns the notification channel for updates.
+	C() <-chan struct{}
+}
+
+// NewUpdatableFn creates an updatable.
+type NewUpdatableFn func() (Updatable, error)
+
+// GetFn returns the latest value.
+type GetFn func(updatable Updatable) (kv.Versionable, error)
```
xichen2020: Would it better to call this `GetUpdateFn`?

placement/staged_placement_watcher.go
xichen2020: Hm why would it depend on `GetFn`? It seems `processFn` is only called from `updateWithLock` in `value.go` and the first line in that function is checking whether the update is nil?

cw9: Because the if the user pass in a GetFn that returns `nil, nil` then the update would be nil

xichen2020: Sure but that should be caught by the nil check on the first line of `updateWithLock` I'd think?

cw9:
Now that I look at it, I think the GetFn and ProcessFn could not use `Versionable` because neither the `Placement` or `Topic` use case have access to the revision field like the `kv.Value` has, and they can not use `Version` to decide whether the update is newer either because that would break when there is a delete. I think my usage of PlacementWatch and TopicWatch are more similar to what the m3x.Watch has, which is to just process whatever the latest value is available on the watch.

placement/staged_placement_watcher.go
cw9: yeah that is true, can remove this

PR #226, opened 2018-05-28
Add ability to mark shards available in batch

robskillington:
When a cluster comes online it needs to mark a lot of shards as available, this adds the ability to mark multiple shards as available in one go to help reduce the RPCs required when a new node with a lot of shards comes online.


placement/algo/sharded_helper.go
```diff
@@ -848,7 +851,7 @@ func markAllShardsAvailable(
 	for _, instance := range p.Instances() {
 		for _, s := range instance.Shards().All() {
 			if s.State() == shard.Initializing {
-				p, err = markShardAvailable(p, instance.ID(), s.ID(), opts)
+				p, err = markShardsAvailable(p, instance.ID(), []uint32{s.ID()}, opts)
```
cw9: can you do something like
```
p, err = markShardsAvailable(p, instance.ID(), instance.Shards().ShardsForState(shard.Initializing), opts) 
```
here and save the iteration?

robskillington: @cw9 hm I tried that but getting some strange test errors.  Can you help me out here (I'm travelling the next day and a bit):
```

prateek: talked offline: going to leave as is because the `markShardsAvailable` method expects `[]uint32` and `ShardsForState(..)` returns `[]Shard`, so we'd have to do the extra iteration to convert anyway.

PR #228, opened 2018-06-21
By default disable etcd with progress notify

cw9:


kv/etcd/options.go
```diff
@@ -69,6 +70,12 @@ type Options interface {
 	// SetWatchChanResetInterval sets the WatchChanResetInterval
 	SetWatchChanResetInterval(t time.Duration) Options
 
+	// WatchWithProgressNotify returns whether to enable WatchWithProgressNotify
+	// on etcd watchers.
+	WatchWithProgressNotify() bool
```
xichen2020: Empty line in between.

kv/etcd/store.go
```diff
@@ -71,17 +71,15 @@ func NewStore(etcdKV clientv3.KV, etcdWatcher clientv3.Watcher, opts Options) (k
 			diskReadError:  scope.Counter("disk-read-error"),
 		},
 	}
+	etcdWatchOpts := []clientv3.OpOption{clientv3.WithCreatedNotify()}
```
xichen2020: Do we need created notify? Should we make this an option as well and disable it by default as well?

PR #231, opened 2018-06-25
[kv] Support watch start revision in watch store

schallert:
Some use cases require specifying the start revision of a watch request.
Since our only use case so far needs the start revision at a store level
as opposed to on a per-watch basis, this PR implements watch revision
support at the store level.

Includes tests to make sure that we can safely specify a start revision
that has been compacted or one that is in the future with no issues.

### DIFF
--- a/etcd/watchmanager/manager.go
+++ b/etcd/watchmanager/manager.go
@@ -152,7 +152,6 @@ func (w *manager) Watch(key string) {
 					w.logger.Warnf("recreating watch at revision %d", r.CompactRevision)
 					startRev = r.CompactRevision
 					watchChan = nil
-					continue
 				}
 			}
 
--- a/etcd/watchmanager/manager.go
+++ b/etcd/watchmanager/manager.go
@@ -28,6 +28,7 @@ import (
 	"github.com/m3db/m3x/log"
 
 	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/etcdserver/api/v3rpc/rpctypes"
 	"github.com/uber-go/tally"
 )
 
@@ -66,17 +67,21 @@ type metrics struct {
 	etcdWatchReset  tally.Counter
 }
 
-func (w *manager) watchChanWithTimeout(key string) (clientv3.WatchChan, context.CancelFunc, error) {
+func (w *manager) watchChanWithTimeout(key string, rev int64) (clientv3.WatchChan, context.CancelFunc, error) {
 	doneCh := make(chan struct{})
 
 	ctx, cancelFn := context.WithCancel(clientv3.WithRequireLeader(context.Background()))
 
 	var watchChan clientv3.WatchChan
 	go func() {
+		wOpts := w.opts.WatchOptions()
+		if rev > 0 {
+			wOpts = append(wOpts, clientv3.WithRev(rev))
+		}
 		watchChan = w.opts.Watcher().Watch(
 			ctx,
 			key,
-			w.opts.WatchOptions()...,
+			wOpts...,
 		)
 		close(doneCh)
 	}()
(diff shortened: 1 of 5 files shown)


kv/etcd/options.go
```diff
@@ -74,6 +74,12 @@ type Options interface {
 	// SetWatchChanInitTimeout sets the WatchChanInitTimeout
 	SetWatchChanInitTimeout(t time.Duration) Options
 
+	// WatchStartRevision is the revision that watch requests will start from.
+	WatchStartRevision() int64
```
xichen2020: Blank line after this line.

kv/etcd/options.go
```diff
@@ -74,6 +74,12 @@ type Options interface {
...
+	// SetWatchStartRevision sets the revision that watch requests will start
+	// from.
+	SetWatchStartRevision(rev int64) Options
```
xichen2020: Wonder if we should name this `SetWatchWithRevision` to be consistent with how this is named in etcd codebase.

kv/etcd/options.go
```diff
@@ -98,6 +104,7 @@ type options struct {
 	watchChanCheckInterval time.Duration
 	watchChanResetInterval time.Duration
 	watchChanInitTimeout   time.Duration
+	watchStartRevision     int64
```
xichen2020: So by default we watch with revision 0 presumably?

kv/etcd/store_test.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/m3db/m3cluster/kv"
 	"github.com/m3db/m3cluster/mocks"
 	"github.com/stretchr/testify/require"
```
xichen2020: Blank line before line 36

kv/etcd/store_test.go
```diff
@@ -780,6 +781,47 @@ func TestTxn_UnknownType(t *testing.T) {
 	require.Equal(t, kv.ErrUnknownCompareType, err)
 }
 
+// Test that watching from 1) an old compacted start revision and 2) a start
+// revision in the future are both safe
+func TestStartRevision(t *testing.T) {
```
xichen2020: nit: `TestWatchWithStartRevision` might be a better name.

kv/etcd/store_test.go
```diff
@@ -780,6 +781,47 @@ func TestTxn_UnknownType(t *testing.T) {
 	require.Equal(t, kv.ErrUnknownCompareType, err)
 }
 
+// Test that watching from 1) an old compacted start revision and 2) a start
```
xichen2020: Function-level comments should start with the name of the function.

kv/etcd/options.go
schallert: We won't even bother setting the value if the user doesn't specify a non-zero start revision. 0 default is what etcd does under the hood as well, if the request has start revision 0 the server just assumes "now"

etcd/watchmanager/manager.go
```diff
@@ -139,6 +145,15 @@ func (w *manager) Watch(key string) {
...
 				// do not stop here, even though the update contains an error
 				// we still take this chance to attempt a Get() for the latest value
+
+				// If the current revision has been compacted, set watchChan to
+				// nil so the watch is recreated with a valid start revision
+				if err == rpctypes.ErrCompacted {
+					w.logger.Warnf("recreating watch at revision %d", r.CompactRevision)
+					startRev = r.CompactRevision
+					watchChan = nil
+					continue
```
cw9: Is this guaranteed to get the watch created at some point? By `continue` here, you basically skipped the `updateFn` logic which would update the kv.Watch with latest value or local cached value when the watch could not be watched, we should probably keep that behavior in case the watcher creation never success. I guess you can get it by removing the continue? Please add a unit test for this case as well.

etcd/watchmanager/manager.go
```diff
@@ -95,14 +100,15 @@ func (w *manager) Watch(key string) {
 	ticker := time.Tick(w.opts.WatchChanCheckInterval()) //nolint: megacheck
 
 	var (
+		startRev  int64
```
cw9: nit: can you call this `revOverride` or something like that? it seems a bit confusing to me looking at this var here while the user already passed in a `withRev`

PR #232, opened 2018-06-26
Add retry options for creating etcd client

robskillington:
When M3 processes using m3cluster for service discovery comes up, a lot of processes will log a fatal error message and fail if they cannot start up.  

This is fine but can cause a large amount of DNS requests to be issued if restarted by a processes supervisor very frequently.  To avoid some of these thundering herd issues if a KV client is down, we add some level of backoff when trying to connect to service discovery.

### DIFF
--- a/client/etcd/options.go
+++ b/client/etcd/options.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/m3db/m3cluster/services"
 	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/retry"
 )
 
 const (
@@ -37,6 +38,12 @@ const (
 	defaultKeepAlivePeriod          = 30 * time.Minute
 	defaultKeepAlivePeriodMaxJitter = 0
 	defaultKeepAliveTimeout         = 20 * time.Second
+
+	defaultRetryInitialBackoff = 2 * time.Second
+	defaultRetryBackoffFactor  = 2.0
+	defaultRetryMaxBackoff     = 4
+	defaultRetryMaxRetries     = 3
+	defaultRetryJitter         = true
 )
 
 type keepAliveOptions struct {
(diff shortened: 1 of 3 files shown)


client/etcd/options.go
```diff
@@ -30,13 +30,20 @@ import (
...
 
 const (
 	defaultKeepAliveEnabled         = false
 	defaultKeepAlivePeriod          = 30 * time.Minute
 	defaultKeepAlivePeriodMaxJitter = 0
 	defaultKeepAliveTimeout         = 20 * time.Second
+
+	defaultRetryInitialBackoff = 2 * time.Second
+	defaultRetryBackoffFactor  = 2.0
+	defaultRetryMaxBackoff     = 4
```
xichen2020: Should this be `4 * time.Second` rather?

client/etcd/options.go
```diff
@@ -159,17 +166,26 @@ func NewOptions() Options {
...
 		sdOpts: services.NewOptions(),
 		iopts:  instrument.NewOptions(),
+		// NB(r): Set some default retry options so changes to retry
+		// option defaults don't change behavior of this client's retry options
+		retryOpts: retry.NewOptions().
+			SetInitialBackoff(defaultRetryInitialBackoff).
+			SetBackoffFactor(defaultRetryBackoffFactor).
+			SetMaxBackoff(defaultRetryMaxBackoff).
+			SetMaxRetries(defaultRetryMaxRetries).
+			SetJitter(defaultRetryJitter),
```
xichen2020: Do we want to retry forever?

client/etcd/options.go
cw9: +1

cw9:
Can we add some details on why we need this in the pr description?

client/etcd/options.go
robskillington: Sure thing, I'll just set this to zero.

client/etcd/options.go
robskillington: Nah, crash looping the process after 16s is fine.  We just don't want to crash loop the process with no backoff at all.

PR #233, opened 2018-07-02
[client] Allow configuration of watch w/ revision

schallert:
This allows `WatchWithRevision` to be set (or not set) on a per-client
basis and overridden on a per-store basis if desired.

### DIFF
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -183,7 +183,7 @@ func (c *csclient) newkvOptions(
 }
 
 // txnGen assumes the caller has validated the options passed if they are
-// user-supplied (as opposed to constructed ourselves)
+// user-supplied (as opposed to constructed ourselves).
 func (c *csclient) txnGen(
 	opts kv.OverrideOptions,
 	cacheFileFn cacheFileForZoneFn,
--- a/client/etcd/client.go
+++ b/client/etcd/client.go
@@ -144,7 +144,7 @@ func (c *csclient) createServices(opts services.OverrideOptions) (services.Servi
 }
 
 func (c *csclient) createTxnStore(opts kv.OverrideOptions) (kv.TxnStore, error) {
-	// validate the override options because they are user supplied
+	// validate the override options because they are user supplied.
 	if err := opts.Validate(); err != nil {
 		return nil, err
 	}
@@ -154,7 +154,7 @@ func (c *csclient) createTxnStore(opts kv.OverrideOptions) (kv.TxnStore, error) 
 func (c *csclient) kvGen(fn cacheFileForZoneFn) services.KVGen {
 	return services.KVGen(func(zone string) (kv.Store, error) {
 		// we don't validate or sanitize the options here because we're using
-		// them as a container for zone
+		// them as a container for zone.
 		opts := kv.NewOverrideOptions().SetZone(zone)
 		return c.txnGen(opts, fn)
 	})
@@ -169,7 +169,7 @@ func (c *csclient) newkvOptions(
 			SetLogger(c.logger).
 			SetMetricsScope(c.kvScope)).
 		SetCacheFileFn(cacheFileFn(opts.Zone())).
-		SetWatchWithRevision(opts.WatchWithRevision())
+		SetWatchWithRevision(c.opts.WatchWithRevision())
 
 	if ns := opts.Namespace(); ns != "" {
 		kvOpts = kvOpts.SetPrefix(kvOpts.ApplyPrefix(ns))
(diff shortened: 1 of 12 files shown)


kv/config.go
```diff
@@ -22,15 +22,17 @@ package kv
 
 // OverrideConfiguration is the config for OverrideOptions.
 type OverrideConfiguration struct {
-	Zone        string `yaml:"zone"`
-	Environment string `yaml:"environment"`
-	Namespace   string `yaml:"namespace"`
+	Zone              string `yaml:"zone"`
+	Environment       string `yaml:"environment"`
+	Namespace         string `yaml:"namespace"`
+	WatchWithRevision int64  `yaml:"watchWithRevision"`
```
cw9: The Override config/options are managed by us and not specific to the data store, but the revision is pretty specific to the data store, for example etcd supports it but many other store don't, can we not have this field here and only keep it in the etcd specific kv options?

schallert: I agree that it's pretty implementation-specific, but I wanted to provide flexibility to set this on a per-store basis instead of having to set it on every store created from a client. If we're fine with this being set at the client level and every store created from the client inheriting it I'll change it. Do you think we need the flexibility or can set this on every store from a single client?

client/etcd/client.go
```diff
@@ -347,6 +354,10 @@ func (c *csclient) sanitizeOptions(opts kv.OverrideOptions) (kv.OverrideOptions,
 		opts = opts.SetEnvironment(c.opts.Env())
 	}
 
+	if opts.WatchWithRevision() == 0 {
```
cw9: should be `if opts.WatchWithRevision() != 0`?

client/etcd/config.go
```diff
@@ -73,12 +73,13 @@ func (c *keepAliveConfig) NewOptions() KeepAliveOptions {
...
-	CacheDir     string                 `yaml:"cacheDir"`
-	ETCDClusters []ClusterConfig        `yaml:"etcdClusters"`
-	SDConfig     services.Configuration `yaml:"m3sd"`
+	Zone              string                 `yaml:"zone"`
+	Env               string                 `yaml:"env"`
+	Service           string                 `yaml:"service" validate:"nonzero"`
+	CacheDir          string                 `yaml:"cacheDir"`
+	ETCDClusters      []ClusterConfig        `yaml:"etcdClusters"`
+	SDConfig          services.Configuration `yaml:"m3sd"`
+	WatchWithRevision int64                  `yaml:"watchWithRevision"`
```
cw9: Do you still need this if you already have it in the per store override options?

client/etcd/client.go
```diff
@@ -144,60 +144,67 @@ func (c *csclient) createServices(opts services.OverrideOptions) (services.Servi
 }
 
 func (c *csclient) createTxnStore(opts kv.OverrideOptions) (kv.TxnStore, error) {
+	// validate the override options because they are user supplied
```
cw9: nit: period at the end of each comment

client/etcd/config.go
schallert: Yeah, some users may not use overrides and may want all stores created from the client to have the revision set. I.e. if the only way I'm consuming my store is by using `NewConfigServiceClient` and then `client.KV()`, I'm not using overrides but I still want the watch revision set.

client/etcd/client.go
schallert: It looks like we do this for function comments but not inline comments (i.e. [here](https://github.com/m3db/m3cluster/blob/d971450316604c151719f53afbb95539e6dbafbb/client/etcd/client.go#L52) and [here](https://github.com/m3db/m3cluster/blob/d971450316604c151719f53afbb95539e6dbafbb/client/etcd/client.go#L311) in this file). I'm inclined to stick to that convention

client/etcd/client.go
schallert: My intention here was that if the override doesn't specify a start revision it will inherit the base config's. This does however mean that if the base config sets `WatchWithRevision`, the override can't disable it. I think that's fine since that's what we do for environment and namespace as well.

client/etcd/client.go
cw9: That was before we started to enforce the period thing, I think going forward for all the new comments we should always add a period.

client/etcd/config.go
cw9: If we are adding the config on the client level, can we only add it here? So that we can still keep the `WatchWithRevision` field to the etcd implementations without polluting the kv.OverrideOptions interface

PR #234, opened 2018-07-02
Add Equal method to ServiceID interface

cw9:


services/services.go
```diff
@@ -809,6 +809,13 @@ func (sid *serviceID) Zone() string                      { return sid.zone }
 func (sid *serviceID) SetName(n string) ServiceID        { sid.name = n; return sid }
 func (sid *serviceID) SetEnvironment(e string) ServiceID { sid.env = e; return sid }
 func (sid *serviceID) SetZone(z string) ServiceID        { sid.zone = z; return sid }
+
+func (sid *serviceID) Equal(other ServiceID) bool {
```
xichen2020: Not sure if you want to protect against nil here.

cw9: sg, will do

PR #236, opened 2018-08-02
[services] Leader: Add read-only Observe API

schallert:
This PR adds a read-only watch-like API to the leader service. This
allows clients that wish to watch an election to do so with without
polling.

Multiple callers can observe an election simultaneously, and when the
client is closed all observers will be gracefully closed.


services/leader/election/client.go
```diff
@@ -180,6 +180,43 @@ func (c *Client) Leader(ctx context.Context) (string, error) {
...
+		for {
+			select {
+			case resp, ok := <-leaderCh:
+				if !ok {
+					close(ch)
+					return
+				}
+
+				for _, v := range resp.Kvs {
+					ch <- string(v.Value)
```
xichen2020: Should this just use the first value?

services/leader/client.go
```diff
@@ -77,10 +79,16 @@ func newClient(cli *clientv3.Client, opts Options, electionID string) (*client,
 		return nil, err
 	}
 
+	// allow multiple observe calls with the same parent context, to be cancelled
```
xichen2020: `Allow ... is closed.`

services/leader/election/client.go
```diff
@@ -203,8 +203,8 @@ func (c *Client) Observe(ctx context.Context) (<-chan string, error) {
 					return
 				}
 
-				for _, v := range resp.Kvs {
-					ch <- string(v.Value)
+				if len(resp.Kvs) > 0 {
+					ch <- string(resp.Kvs[0].Value)
```
xichen2020: nit: add a comment around why only the first value is taken, possibly referring to the etcd code.

PR #240, opened 2018-10-04
[ci] Use buildkite for CI

schallert:


.buildkite/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -ex
+
```
prateek: Don’t need this. The CI does it as part of pulling the repo

schallert: Hmm I removed this and don't see the steps in the output anymore. Going to re-add

prateek: lol meant the submodule recurse init thing, all good.

PR #245, opened 2018-10-12
Refresh cache on key deletion

prateek:


kv/etcd/store.go
```diff
@@ -389,10 +389,11 @@ func (c *client) getFromKVStore(key string) (kv.Value, error) {
 func (c *client) getFromEtcdEvents(key string, events []*clientv3.Event) kv.Value {
 	lastEvent := events[len(events)-1]
 	if lastEvent.Type == clientv3.EventTypeDelete {
+		c.deleteCache(key)
```
robskillington: This was the fix? cc @cw9

schallert:
Will this protect us against the case where we miss the delete event? I have to look at the code to see how we replay events, but even if we do I want to make sure we're protected against the following series of events:

1. Client gets value `v1` for key `k`, and puts that in the cache
2. Client shuts down
3. `k` is deleted from etcd
4. etcd runs a compaction to the latest revision, thus deleting the fact that key ever existed and can't replay the delete event
5. Client comes back up

kv/etcd/store.go
prateek: Aye. We handled cache eviction correctly for deletes in the client doing the delete. But did not handle it correctly for anyone watching on the key. Without this change, on receiving an update we early terminated without updating the cache.

prateek:
@schallert fair question. I’ll add a test to ensure we handle the situation correctly.

integration/cache_deletion_test.go
```diff
@@ -0,0 +1,181 @@
...
+	etcdclient "github.com/m3db/m3cluster/client/etcd"
+	"github.com/m3db/m3cluster/generated/proto/commonpb"
+	"github.com/m3db/m3cluster/integration/etcd"
+	xclock "github.com/m3db/m3x/clock"
+
+	"github.com/coreos/pkg/capnslog"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCacheDeleteInvalidation(t *testing.T) {
```
cw9: Can this just be a unit test? There are a few cache related unit test in store_test.go already, would be nice to keep them together

kv/etcd/store.go
cw9: I vaguely recall that we used to specifically not delete cache on watch updates, to prevent mistakenly deleting a key (e.g placement). This will change that assumption, I'm not sure if there are other use cases in m3db that relies on that assumption.

cw9: Also to for the case Matt mentioned, you might need to call `deleteCache` on line 406 as well

prateek: re: not deleting - after the stale cache issue Matt found in prod, i think we should add this change for safety. if m3db/other clients needs not apply deletes they should do so in the application layer (e.g. what we do for namespace removes - we see the update and choose not to act on it).

prateek: re: extra deleteCache call, cheers - that helps narrow my spelunking :)

integration/cache_deletion_test.go
prateek: Sure thing, didn't realise that it uses the integration package too.

prateek: done

kv/etcd/store.go
prateek: @cw9 interestingly found something else that i don't understand. `initCache` decodes the local cache file, but if it succeeds at that, it doesn't load the cache with those values. What was the intent there? 

see test: `TestStaleDelete` to see what I mean.

kv/etcd/store_test.go
```diff
@@ -653,6 +735,85 @@ func TestDelete_TriggerWatch(t *testing.T) {
...
+	require.True(t, xclock.WaitUntil(func() bool {
+		return 1 == len(getClient.cache.Values)
+	}, 2*time.Second), "timed out waiting for client to read values from cache")
+	println(fmt.Sprintf("%+v", getClient.cache.Values))
+
+	// get value and ensure it's able to serve the read
+	v, err := getClient.Get("foo")
+	require.NoError(t, kv.ErrNotFound)
+	require.NotNil(t, v)
+	// TODO: should this not serve the read? it's got the value in the cache and can't talk to etcd.
```
prateek: while trying to repro matt's case i ended up here: it appears that this isn't serving a read from the cache while etcd is down. 

i'm probably missing something atm but reminder for us to talk about it @cw9

kv/etcd/store_test.go
```diff
@@ -653,6 +735,85 @@ func TestDelete_TriggerWatch(t *testing.T) {
...
+	getClient := getStore.(*client)
+
+	require.True(t, xclock.WaitUntil(func() bool {
+		return 1 == len(getClient.cache.Values)
+	}, 2*time.Second), "timed out waiting for client to read values from cache")
+	println(fmt.Sprintf("%+v", getClient.cache.Values))
+
+	// get value and ensure it's able to serve the read
+	v, err := getClient.Get("foo")
+	require.NoError(t, kv.ErrNotFound)
```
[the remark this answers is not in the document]
prateek: lol good call.

kv/etcd/store_test.go
```diff
@@ -653,6 +735,85 @@ func TestDelete_TriggerWatch(t *testing.T) {
...
+	// make a new etcd cluster (to mimic the case where etcd has deleted and compacted
+	// the value).
+	newServerCachePath, err := ioutil.TempDir("", "new-server-cache-dir")
+	require.NoError(t, err)
+	defer os.RemoveAll(newServerCachePath)
+
+	// write the cache file with correct contents in the new location
+	f, err := os.Create(path.Join(newServerCachePath, fileName))
+	require.NoError(t, err)
+	_, err = f.Write(cacheBytes)
```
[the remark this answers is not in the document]
prateek: So I do that below –- check the WaitUntil on line 789. It verifies content is the same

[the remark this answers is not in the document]
cw9: They could be equally empty?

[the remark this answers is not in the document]
cw9: scratch that, there is a check later to make sure there are values in new cache.

